PR #1, opened 2017-03-01
Add user accessor to AlexaEvent

heneryville:


armonge:
We also need to update the documentation on this

PR #2, opened 2017-03-02
Adding travisci and coveralls 

armonge:


run-ci.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -ev
+
+npm run test-ci
+npm run cobertura
+npm run lint
```
chrux: Shouldn't we add his 3 libs as devDependencies?

armonge: @chrux only new dependency is coveralls, the rest are already in package.json devDependencies.

I don't want coveralls in there cause it's only ever used when this is run in CI

PR #8, opened 2017-03-06
Update README.md

armonge:


rmberrios:
@armonge will merge,
BTW, it says the group is invite only, Are you going to leave it like that?

armonge:
@rmberrios my bad, group is Voxa-framework, which should be public

PR #14, opened 2017-03-13
Session attributes are now only taken from serialize

### LINKED ISSUE #10: Session attributes are aggresively persistant
Our handling of session attributes is make it practically impossible to remove stuff from the session. In StateMachineSkill line 53 (the onBeforeReplySent handler), the old session, and the result of the data serialize are *merged* together.

I'm mostly OK with the idea that you've got to delete things explicitly from the session to make them not persist, (though my functional part of me hates this), but that data values from prior request are allowed to clober the next request is not OK. Voxa should enforce that everything the model wants to go to the next session will be in serialize, and it should not merge in the prior data attributes. This is why the original statemachine code used Object.assign here, rather than _.merge.

armonge:
Fixes #10

### DIFF
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -8,7 +8,7 @@ class Model {
   }
 
   static fromEvent(alexaEvent) {
-    return new this(alexaEvent.session.attributes.data);
+    return new this(alexaEvent.session.attributes.modelData);
   }
 
   serialize() {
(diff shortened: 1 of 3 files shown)


heneryville:
So.. I'm not sure that this is the correct solution either. In this world plugins would never be able to persist any information from request-to-request; only things with access to the model and how it serializes.

We should talk about what strategy we use here. One idea is to allow the session the have (sticky attributes), but just not for the data attribute, which is entirely controlled by the model and is NOT merged.

armonge:
plugins can always hook into `onBeforeReplySent` and add things to `session.attributes`. However, `session.attributes.data` seems like it should "only" be handled by the model, i can see a situation were we have issues with plugins overwriting what's in `session.attributes.data`

heneryville:
@armonge Approved. It's just out of date now.

PR #16, opened 2017-03-14
Multiple directives support

armonge:


lib/Reply.js
```diff
@@ -72,7 +72,13 @@ class Reply {
     this.msg.card = msg.card || this.msg.card;
     this.msg.yield = this.msg.yield || !!(msg.ask || msg.tell);
     this.msg.hasAnAsk = this.msg.hasAnAsk || !!msg.ask;
-    this.msg.directives = msg.directives || this.msg.directives;
+    if (msg.directives) {
+      if (_.isArray(msg.directives)) {
+        this.msg.directives = _.concat(this.msg.directives, msg.directives);
+      } else {
+        this.msg.directives.push(msg.directives);
```
heneryville: I love that you're backwards supporting here.

lib/Reply.js
```diff
@@ -84,6 +90,9 @@ class Reply {
     this.msg.reprompt = reply.msg.reprompt || this.msg.reprompt;
     this.msg.card = reply.msg.card || this.msg.card;
     this.msg.directives = reply.msg.directives || this.msg.directives;
+    if (reply.msg.directives) {
+      this.msg.directives = _.concat(this.msg.directives, reply.msg.directives);
```
heneryville: Seems like this combined with the line above it could have some unintended consequences. I'd recommend just dropping the above line and the if statement?

lib/Reply.js
```diff
@@ -150,28 +159,25 @@ class Reply {
...
+        .map((directive) => {
+          if (directive.playBehavior) {
+            return {
+              type: directive.type,
+              playBehavior: directive.playBehavior,
+              audioItem: {
+                stream: {
+                  token: directive.token,
+                  url: directive.url,
+                  offsetInMilliseconds: directive.offsetInMilliseconds,
```
heneryville: So... we're going to keep the mapping that exists right now?

Issue #17, opened 2017-03-14
state-flow plugin crashes when doing fallback


heneryville:
The state-flow plugin crashes when there is no fallback. It fails here:

```
  skill.onAfterStateChanged((alexaEvent, reply, transition) => {
    alexaEvent.flow.push(transition.to);
  });
```

Should onAfterStateChanged be called if we're in fallback? If so, we need to put loud notes in docs that transition could be null. However, we've already got onUnhandledState, so I'm not sure this is necessary.

PR #18, opened 2017-03-14
Fix for state-flow plugin

armonge:
Also adding the alexaEvent to the Reply object, this just makes it
easier to test assertions

### DIFF
--- a/lib/plugins/state-flow.js
+++ b/lib/plugins/state-flow.js
@@ -7,9 +7,7 @@ function register(skill) {
   });
 
   skill.onAfterStateChanged((alexaEvent, reply, transition) => {
-    if (transition) { // unhandled state
-      alexaEvent.flow.push(transition.to);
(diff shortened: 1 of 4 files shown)


armonge:
This addresses #17

lib/plugins/state-flow.js
```diff
@@ -7,7 +7,9 @@ function register(skill) {
   });
 
   skill.onAfterStateChanged((alexaEvent, reply, transition) => {
-    alexaEvent.flow.push(transition.to);
+    if (transition) { // unhandled state
```
heneryville: So I get that we're null-checking this now, but I'm not sure if onAfterStateChanged should even be called if there is no transition. Isn't this what onUnhandled is for? If we do choose to make transition sometimes null, we need to make that loud and clear in the docs, and explain what circumstances make it null.

PR #19, opened 2017-03-14
Cloudwatch Plugin should report the absense of an error for statistical purposes

cvaca7:
Send metric data to cloud-watch plugin every time that onBeforeReplySent is executed


lib/plugins/cloud-watch.js
```diff
@@ -15,17 +15,22 @@ function register(skill, cloudwatch, eventMetric) {
     ],
     Namespace: 'StateMachine',
```
armonge: let's provide more information on which parameters "should" be  overriden in the documentation. For example, the documentation provides an example where Namespace is overwritten, but there's no explanation on why that is. the rationale is that you want a separate namespace per skill

lib/plugins/cloud-watch.js
```diff
@@ -15,17 +15,22 @@ function register(skill, cloudwatch, eventMetric) {
...
   };
+  const eventMetricConfig = _.merge(defaultEventMetric, eventMetric);
+  const putMetricDataAsync = Promise.promisify(cloudwatch.putMetricData);
+  let positiveEvenMetric = defaultEventMetric;
```
armonge: i'm not sure what's the difference between eventMetricConfig and positiveEventMetric, you're basically tracking the same thing in `onBeforeReplySent` and `onStateMachineError`

test/plugins/cloud-watch.spec.js
```diff
@@ -14,6 +14,43 @@ const views = require('../views');
...
+
+    return stateMachineSkill.execute(event)
+      .then((reply) => {
+        expect(reply.msg.statements[0]).not.equal('An unrecoverable error occurred.');
+        expect(cloudwatchMock.called).to.be.true;
+        expect(cloudwatchMock.callCount).to.be.at.most(1);
+        expect(cloudwatchMock.lastCall.args[0].Namespace).to.equal('fooBarSkill');
+        expect(cloudwatchMock.lastCall.args[0].MetricData[0].MetricName).to.equal('Caught Error');
+        expect(cloudwatchMock.lastCall.args[0].MetricData[0].Unit).to.equal('Count');
+        expect(cloudwatchMock.lastCall.args[0].MetricData[0].Value).to.equal(1);
```
armonge: this should be equal to 0, you wat to track sucessful invocations as 0

lib/plugins/cloud-watch.js
```diff
@@ -16,21 +21,54 @@ function register(skill, cloudwatch, eventMetric) {
...
+  }));
 
   skill.onError(alexaEvent => Promise.try(() => {
     const errorReported = _.get(alexaEvent, 'cloudwatch.errorReported');
 
     if (!errorReported) {
-      return putMetricDataAsync(eventMetricConfig).then(() => null);
+      _.set(onErrorMetric, 'MetricData[0].Timestamp', new Date());
+      return putMetricDataAsync(onErrorMetric).then((data, error) => {
+        debug(`onError Metric Data Error: ${JSON.stringify(error)}`);
```
armonge: not sure what's the error parameter in here, what do the logs say when executing this?

test/plugins/cloud-watch.spec.js
```diff
@@ -14,6 +14,43 @@ const views = require('../views');
...
+      },
+    };
+
+    cloudwatchPlugin(stateMachineSkill, cloudwatch, eventMetric);
+
+    stateMachineSkill.onIntent('SomeIntent', () => ({}));
+
+    return stateMachineSkill.execute(event)
+      .then((reply) => {
+        expect(reply.msg.statements[0]).not.equal('An unrecoverable error occurred.');
```
armonge: should just test for `reply.error` to be undefined

lib/plugins/cloud-watch.js
```diff
@@ -16,21 +21,54 @@ function register(skill, cloudwatch, eventMetric) {
...
+  }));
+
+  skill.onStateMachineError(alexaEvent => Promise.try(() => {
     _.set(alexaEvent, 'cloudwatch.errorReported', true);
-    return null;
-  })));
+    _.set(onErrorMetric, 'MetricData[0].Timestamp', new Date());
 
+    return putMetricDataAsync(onErrorMetric).then((data, error) => {
+      debug(`onStateMachineError Metric Data Error: ${JSON.stringify(error)}`);
```
armonge: what is error?

cvaca7: The intention is that in case that the metric data couldn't be sent, tell the developer whats wrong.

armonge:
@cvaca7 also seems like coverage fell on this file https://coveralls.io/builds/10629011/source?filename=lib%2Fplugins%2Fcloud-watch.js

lib/plugins/cloud-watch.js
armonge: errors in promises don't work like that, check http://bluebirdjs.com/docs/api/then.html

`then` supports a separate error handler function, not a second `error` parameter, also, in general is better to attach a `catch` handler instead of having the second function in `then`

PR #21, opened 2017-03-14
Autoload plugin

omenocal:


lib/plugins/auto-load.js
```diff
@@ -0,0 +1,40 @@
...
+    if (!config) {
+      config = { loadByToken: false };
+    }
+
+    if (config.loadByToken && !_.get(alexaEvent, 'session.user.accessToken')) {
+      debug('Missing token on load data by token operation');
+      return alexaEvent;
+    }
+
+    let key = _.get(alexaEvent, 'session.user.userId') || _.get(alexaEvent, 'lambdaContext.System.user.userId');
```
armonge: you should just be using http://voxa.readthedocs.io/en/latest/alexa-event.html#AlexaEvent.user

Also, `lambdaContext` has the information from `context` in the `(event, context, callback)` lambda handler, The context from the handler event is in 'alexaEvent.context.System.userId.

lib/plugins/auto-load.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+const _ = require('lodash');
+const debug = require('debug')('voxa');
+
+function autoLoad(skill, adapter, config) {
```
armonge: i think i'd be better if we just had a `config` object with an `adapter` key, that's what we have in all other plugins

package.json
```diff
@@ -38,6 +38,7 @@
   "devDependencies": {
     "chai": "^3.5.0",
     "chai-as-promised": "^6.0.0",
+    "dynasty": "^0.2.7",
```
armonge: this adds a new requirement, at this point i think we want a new package.

lib/plugins/auto-load.js
```diff
@@ -0,0 +1,40 @@
...
+    let key = _.get(alexaEvent, 'session.user.userId') || _.get(alexaEvent, 'lambdaContext.System.user.userId');
+
+    if (config.loadByToken) {
+      key = alexaEvent.session.user.accessToken;
+    }
+
+    if (typeof adapter.get === 'function') {
+      return adapter.get(key)
+      .then((data) => {
+        _.assign(alexaEvent.model, data);
```
armonge: this gets information from the user? if so why are we overwriting the whole model instead of just `alexaEvent.model.user = data`

lib/plugins/auto-load.js
```diff
@@ -0,0 +1,40 @@
...
+        return alexaEvent;
+      })
+      .catch((error) => {
+        debug(`Error getting data: ${error}`);
+        return alexaEvent;
```
armonge: do we really want to silence this errors?

test/autoLoadAdapter.js
```diff
@@ -0,0 +1,44 @@
...
+ * Copyright (c) 2016 Rain Agency.
+ * Licensed under the MIT license.
+ */
+
+const Dynasty = require('dynasty');
+const https = require('https');
+
+class AutoLoadAdapter {
+  constructor() {
+    const dynasty = Dynasty({
```
armonge: doesn't dynasty require the aws credentials? that's the reason i'm using a different ORM in https://github.com/mediarain/voxa/blob/master/samples/accountLinking/services/userStorage.js#L12

lib/plugins/auto-load.js
```diff
@@ -0,0 +1,40 @@
...
+      });
+    }
+
+    debug('No get method to fetch data from');
```
armonge: this should raise an exception on starting the plugin, not silently fail on runtime

test/autoLoadAdapter.js
```diff
@@ -0,0 +1,44 @@
...
+        agent: new https.Agent({
+          keepAlive: false,
+          rejectUnauthorized: true,
+          secureProtocol: 'TLSv1_method',
+          ciphers: 'ALL',
+        }),
+      },
+    });
+
+    this.userTable = dynasty.table('FakeTable');
```
armonge: the table should be configurable

docs/plugins.rst
```diff
@@ -125,3 +125,31 @@ How to use it
...
+
+Params
+******
+
+.. js:function:: autoLoad(skill, adapter, [config])
+
+  Autoload plugin uses ``skill.onRequestStarted`` to load data the first time user open a skill
+
+  :param Voxa skill: The stateMachineSkill.
+  :param adapter: Any object with a ``get`` method to fetch information from database.
```
armonge: the signature of the function changed

lib/plugins/auto-load.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+const _ = require('lodash');
+const debug = require('debug')('voxa');
+
+function autoLoad(skill, config) {
+  skill.onRequestStarted((alexaEvent) => {
+    if (!config) {
```
armonge: this and all checks on the config should be right in the autoload function and not inside the onRequestStarted callbacks so it's a configuration error catched on starting the skill and not in runtime

test/autoLoadAdapter.js
```diff
@@ -0,0 +1,47 @@
...
+ * Auto load adapter for tests
+ *
+ * Copyright (c) 2016 Rain Agency.
+ * Licensed under the MIT license.
+ */
+
+const AWS = require('aws-sdk');
+const DOC = require('dynamodb-doc');
+
+class AutoLoadAdapter {
```
armonge: if we're only using this for tests then there's no need to have it actually work with dynamodb and then mock it, let's just have a simpler plain object returning promises for get and put

test/plugins/auto-load.spec.js
```diff
@@ -0,0 +1,139 @@
...
+          userId: 'user-xyz',
+        },
+        new: true,
+      },
+      request: {
+        type: 'LaunchRequest',
+      },
+    };
+
+    simple.mock(AutoLoadAdapter.prototype, 'get')
```
armonge: just what i was saying, we're "never" using the dynamodb getItem implementation, we're always mocking it. Let's remove the dynamodb dependency OR make this a separate package and actually provide the adapter for users so they can use it

samples/accountLinking/skill/MainStateMachine.js
```diff
@@ -5,7 +5,10 @@ const Voxa = require('voxa');
 const views = require('./views');
 const variables = require('./variables');
 const states = require('./states');
+const UserStorage = require('../services/userStorage');
+const userAdapter = new UserStorage();
 
 const skill = new Voxa({ variables, views });
 states.register(skill);
+Voxa.plugins.autoLoad(skill, userAdapter);
```
armonge: :+1: for using the plugins in the samples

armonge:
@omenocal also some conflicts

lib/plugins/auto-load.js
```diff
@@ -3,40 +3,44 @@
...
+  skill.onRequestStarted((alexaEvent) => {
+    if (defaultConfig.loadByToken && !_.get(alexaEvent, 'session.user.accessToken')) {
       debug('Missing token on load data by token operation');
       return alexaEvent;
     }
 
     let key = _.get(alexaEvent, 'user.userId');
 
-    if (config.loadByToken) {
+    if (defaultConfig.loadByToken) {
```
armonge: this should probably just be a part of the adapter, not the plugin. just pass alexaEvent.user to the adapter and have the adapter figure out what to do

test/plugins/auto-load.spec.js
```diff
@@ -121,19 +121,15 @@ describe('AutoLoad plugin', () => {
...
     const skill = new StateMachineSkill({ variables, views });
-    autoLoad(skill, { adapter });
-
     const spy = simple.spy(() => ({ reply: 'LaunchIntent.OpenResponse' }));
-    skill.onIntent('LaunchIntent', spy);
 
-    skill.onError((alexaEvent, error) => {
+    try {
+      autoLoad(skill, { adapter });
+    } catch (error) {
```
armonge: would be better to use `.throw(constructor)`, check http://chaijs.com/api/bdd/

package.json
```diff
@@ -38,6 +38,7 @@
   "devDependencies": {
     "chai": "^3.5.0",
     "chai-as-promised": "^6.0.0",
+    "dynamodb-doc": "^1.0.0",
```
armonge: seems like we're not using this anymore

test/plugins/auto-load.spec.js
```diff
@@ -117,6 +117,38 @@ describe('AutoLoad plugin', () => {
...
+    const skill = new StateMachineSkill({ variables, views });
+    const spy = simple.spy(() => ({ reply: 'LaunchIntent.OpenResponse' }));
+
+    try {
+      autoLoad(skill, { loadByToken: true });
+    } catch (error) {
+      expect(spy.called).to.be.false;
+      expect(spy.lastCall.args).to.be.empty;
+      expect(error).to.be.ok;
+      expect(error.message).to.equal('Empty adapter');
```
armonge: Error message should be more like "missing adapter"

test/plugins/auto-load.spec.js
```diff
@@ -117,6 +117,38 @@ describe('AutoLoad plugin', () => {
...
+    } catch (error) {
+      expect(spy.called).to.be.false;
+      expect(spy.lastCall.args).to.be.empty;
+      expect(error).to.be.ok;
+      expect(error.message).to.equal('Empty config file');
+    }
+  });
+
+  it('should throw an error when no adapter is set up in the config file', () => {
+    simple.mock(adapter, 'get', undefined);
```
armonge: what is this mock for?

test/plugins/auto-load.spec.js
```diff
@@ -117,6 +117,38 @@ describe('AutoLoad plugin', () => {
     return skill.execute(event);
   });
 
+  it('should throw an error when no config file is provided', () => {
+    simple.mock(adapter, 'get', undefined);
```
armonge: what is this mock for?

lib/plugins/auto-load.js
omenocal: Remember other programers will have to code the adapter, and the most straightforward the better, that's why I just pass the actual key to the GET method. As a programmer I wouldn't expect a model.user object to then validate that. I would just like to have available the key to fetch with. Thoughts on that approach?

test/plugins/auto-load.spec.js
omenocal: The `autoLoadAdapter` adapter has a valid get method. This test is to verify if the adapter users would pass through does not have a valid GET method, so we would throw an error for that

lib/plugins/auto-load.js
armonge: Well, the [alexa.user](http://voxa.readthedocs.io/en/latest/alexa-event.html)  is a documented interface and seems to me it's simple enough to just use it. 

I think that having an extra configuration key to decide what key to use doesn't really save you that much when you actually go and create a the adapter.  thoughts on this @rmberrios @heneryville ?

lib/plugins/auto-load.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  if (!_.isFunction(config.adapter.get)) {
+    throw Error('No get method to fetch data from');
+  }
+
+  defaultConfig = _.merge(defaultConfig, config);
+
+  skill.onRequestStarted((alexaEvent) => {
+    if (defaultConfig.loadByToken && !_.get(alexaEvent, 'session.user.accessToken')) {
```
armonge: also, let's get the accessToken from `alexaEvent.user.accessToken`

armonge:
@omenocal let's check the build, seems like a simple lint issue

lib/plugins/auto-load.js
rmberrios: i think the adapter should handle this.

lib/plugins/auto-load.js
```diff
@@ -3,15 +3,15 @@
 const _ = require('lodash');
 const debug = require('debug')('voxa');
 
-let defaultConfig = { loadByToken: false };
+let defaultConfig = {};
 
 function autoLoad(skill, config) {
   if (!config) {
-    throw Error('Empty config file');
+    throw Error('Missing config file');
```
armonge: also, this is not a `file`

docs/plugins.rst
```diff
@@ -128,3 +128,30 @@ How to use it
...
+
+
+Usage
+******
+
+.. code-block:: javascript
+
+    const skill = new Voxa({ Model, variables, views });
+
+    Voxa.plugins.autoLoad(skill, { adapter, loadByToken: false });
```
armonge: loadByToken not needed anymore

samples/accountLinking/skill/states.js
```diff
@@ -1,29 +1,15 @@
 'use strict';
 
-const UserStorage = require('../services/userStorage');
+const _ = require('lodash');
 
 exports.register = function register(skill) {
   skill.onIntent('LaunchIntent', (alexaEvent) => {
-    if (alexaEvent.model.user && alexaEvent.model.user.id) {
+    if (_.get(alexaEvent, 'model.email')) {
```
armonge: we probably need to update the userStorage now

samples/accountLinking/skill/variables.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
 exports.user = function user(model) {
-  if (model.user) {
-    return model.user.email;
+  if (model.email) {
+    return model.email;
```
armonge: is this correct?  seems like https://github.com/mediarain/voxa/pull/21/files#diff-b0d12e95adc3ef86d1cf02fa31ad2e56R25 would set the email in the `model.user.email` key

docs/plugins.rst
```diff
@@ -128,3 +128,30 @@ How to use it
...
+------------------
+
+It accepts an adapter to autoload info into the model object coming in every alexa request.
+
+Params
+******
+
+.. js:function:: autoLoad(skill, [config])
+
+  Autoload plugin uses ``skill.onRequestStarted`` to load data the first time user open a skill
```
armonge: actually, using onRequestStarted means the data is loaded every time the user makes a requests to alexa, not only the first time when the user starts the session

test/plugins/auto-load.spec.js
```diff
@@ -0,0 +1,103 @@
...
+    const skill = new StateMachineSkill({ variables, views });
+    autoLoad(skill, { adapter });
+
+    const spy = simple.spy(() => ({ reply: 'LaunchIntent.OpenResponse' }));
+    skill.onIntent('LaunchIntent', spy);
+
+    simple.mock(adapter, 'get')
+    .rejectWith(new Error('Random error'));
+
+    skill.onError((alexaEvent, error) => {
```
armonge: you should probably assert that your `onError` callback is being called after your execute

omenocal: I just use the `skill.onError` to assert the other things I am testing, not to handle any other special case, is it still worthy to assert it ?

armonge: if all you're asserting is that the session is empty you should better do it like this
```
skill.execute(event)
.then((reply) => {
   expect(reply.error).to.not.be.undefined;
   expect(reply.alexaEvent.session).to.be.undefined;
  expect(reply.error.message).to.equal('Random error');
});
```

But really, i'm not sure what the actual test here is, "should not get data from adapter when adapter throws error on getting data", what does it mean?

omenocal: I will rename it to `should throw error on getting data from adapter`. It's all what I'm asserting, that the mockable method throw an error and the framework to get it

PR #27, opened 2017-03-27
Not crashing when developers do something like `event.model = null`

armonge:


lib/StateMachineSkill.js
```diff
@@ -48,15 +48,28 @@ class StateMachineSkill extends AlexaSkill {
...
-      .then((modelData) => {
-        alexaEvent.session.attributes.modelData = modelData;
+    this.onBeforeReplySent((alexaEvent, reply, transition) => {
+      const serialize = _.get(alexaEvent, 'model.serialize');
+
+      // we do require models to have a serialize method and check that when Voxa is initialized,
+      // however, developers could do stuff like `alexaEvent.model = null`, which seems natural if they want to
+      // clear the model
+      if (!serialize) {
+        alexaEvent.session.attributes.modelData = {};
```
heneryville: Why an empty object rather than null?

lib/StateMachineSkill.js
```diff
@@ -48,15 +48,28 @@ class StateMachineSkill extends AlexaSkill {
...
-          alexaEvent.session.attributes.reply = _.pickBy({ msgPath: transition.reply, state: transition.to.name });
-        }
-      }));
+      return Promise.resolve(alexaEvent.model.serialize())
+        .then((modelData) => {
+          alexaEvent.session.attributes.modelData = modelData;
+          alexaEvent.session.attributes.state = transition.to.name;
+
+          if (reply.msg.hasAnAsk) {
+            alexaEvent.session.attributes.reply = _.pickBy({ msgPath: transition.reply, state: transition.to.name });
```
heneryville: Would we not track this also if there is no serialize function? Also, do we need to track this here? Isn't this what the reprompt-on-bad-response plugin does?

armonge: actually yes. This is leftover from the first state machine codebase, i'm removing it. It's not documented anyway

PR #28, opened 2017-03-28
Voxa 2.1.1 should now work with node 6.10

armonge:


package.json
```diff
@@ -23,7 +23,8 @@
   ],
   "author": "Rain Agency <npm@rain.agency> (http://rain.agency)",
   "engines": {
-    "node": "4.3.2"
+    "node": "4.3.2",
+    "node": "6.10"
```
heneryville: Is this even legitimate JSON? The "node" key is present twice.

package.json
```diff
@@ -23,8 +23,7 @@
   ],
   "author": "Rain Agency <npm@rain.agency> (http://rain.agency)",
   "engines": {
-    "node": "4.3.2",
-    "node": "6.10"
+    "node": ">=4.3.2 <=6.10"
```
heneryville: Do we really need to restrict the maximum engine? I don't think we're that likely to break on higher versions.

PR #33, opened 2017-04-01
Changes required to run starter kit

spatical:
I had to make these changes in order to run the starter kit. 

* Added config to server.js
* In order to set up the test in the alexa developer site, at least one sample utterance is required.


spatical:
@armonge Is there another way Voxa automatically gets the port number?    Or is getting it from the config file like I did here expected?

PR #35, opened 2017-04-03
Need to have a default port for the server

cvaca7:
Default is going to be port 3000


test/AlexaSkill.spec.js
```diff
@@ -11,6 +11,11 @@ const simple = require('simple-mock');
 const _ = require('lodash');
 
 describe('AlexaSkill', () => {
+  it('Skill should able to run without any port as a parameter', () => {
```
armonge: this is not really testing anything, we should either remove the test or rewrite it so it actually makes an assert

armonge:
@heneryville can you take a look at this one, i'm more inclined to delete the test than to keep it as is, not sure if it adds a lot of value and i'm not sure if there's a better way to test it

heneryville:
@cvaca7 Let's drop the unit test altogether. It was a great idea/way to test this, but in the end we're not sure it makes sense to test this particular part.

PR #37, opened 2017-04-17
Returning a better error message for unhandled intents

### LINKED ISSUE #29: Voxa should give a better error when an intent is unhandled
Currently I get:
UnhandledState: Transition from entry resulted in null
    at /Users/mitchellharris/src/alexa/alexa-podcasts/node_modules/voxa/lib/StateMachine.js:40:50
    at tryCatcher (/Users/mitchellharris/src/alexa/alexa-podcasts/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/Users/mitchellharris/src/alexa/alexa-
etc

I'd expect something like, "XXX went unhandled"

armonge:
Fixes #29


lib/Errors.js
```diff
@@ -1,9 +1,10 @@
 'use strict';
 
 class UnhandledState extends Error {
-  constructor(transition, fromState) {
-    const message = `Transition from ${fromState} resulted in ${transition}`;
+  constructor(alexaEvent, transition, fromState) {
+    const message = `${alexaEvent.intent.name} went unhandled`;
```
rmberrios: @armonge Could we add the previous state ? It could be very helpful if is not coming from `entry`

armonge: can you work on that one?

rmberrios: Okay i will! 😄

armonge:
@rommelberrios please update the test

PR #39, opened 2017-04-17
Tutorial leve documentation

cvaca7:
- MVC description diagram for Voxa with responsibilities
- New Alexa developers starting with Voxa
- glossary/index


armonge:
@cvaca7 when creating a new branch always start from the remote master, that way you don't get the huge history, for example in this case it shows as if there were 35 commits on this PR

docs/glossary.rst
```diff
@@ -0,0 +1,60 @@
+Voxa Documentation!
```
armonge: i'm not sure what's this file for, it's practically the same as index right?

docs/new-user.rst
```diff
@@ -0,0 +1,20 @@
...
+AWS Developer account
+---------------------
+
+The first thing you need to do is configure your AWS developer account, `here <http://docs.aws.amazon.com/lambda/latest/dg/setting-up.html>`_ you can see the instructions step by step
+
+
+AWS Lambda function for a custom skill
+--------------------------------------
+
+You need to create a lambda function in order to have your skill up and running, `here you can see how <https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/developing-an-alexa-skill-as-a-lambda-function>`_
```
armonge: Actually you don't, we tend to do all our skill development using ngrok and a local server. You only need a lambda when you're about to publish it. Also, you `could` not use lambda at all, and host your skill in EC2

docs/new-user.rst
```diff
@@ -0,0 +1,20 @@
+New Alexa developer
+==============================================
+
+If the skills development for alexa is a new thing for you, we have some suggestion to get you deep into this world.
+
+AWS Developer account
+---------------------
+
+The first thing you need to do is configure your AWS developer account, `here <http://docs.aws.amazon.com/lambda/latest/dg/setting-up.html>`_ you can see the instructions step by step
```
armonge: you need an amazon developer account, AWS is another thing

docs/new-user.rst
```diff
@@ -0,0 +1,20 @@
...
+
+AWS Lambda function for a custom skill
+--------------------------------------
+
+You need to create a lambda function in order to have your skill up and running, `here you can see how <https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/developing-an-alexa-skill-as-a-lambda-function>`_ 
+
+
+Registering an Alexa skill
+--------------------------
+To configure a new skill, you need an account on the AWS developer portal. If you don't already have an account, go to `aws-developer-account <https://developer.amazon.com/login.html>`_ and create an account. After that, you can follow `these instructions <https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/registering-and-managing-alexa-skills-in-the-developer-portal#registering-an-alexa-skill>`_ to register your skill
```
armonge: you go to the Amazon developer portal, not AWS, and this is actually the first thing you'll need

docs/glossary.rst
cvaca7: Yes, Mitchel wanted to have the index on the glossary. If i just added this index to the glossary it shows twice every item

PR #48, opened 2017-05-29
Fixed issue when adding directives to I18Renderer class

omenocal:


lib/renderers/I18NRenderer.js
```diff
@@ -30,10 +30,14 @@ class I18NRenderer extends DefaultRenderer {
           returnObjects: true,
         });
 
+        if (_.isString(message) && message === result.reply) {
+          throw new Error(`Locale views is not implemented: ${alexaEvent.request.locale}`);
```
armonge: do we really want to crash in here or do we want the skill to simply fallback to the default language strings? Also, seems like the message is not clear, should be something like `Views for en-gb locale are missing`

omenocal: I don't think there should be a default locale since the words are different. It's true US and UK are similar but what if the user program the german version, we shouldn't speak in English looking for a default language.

I can change the error message to this other one.

armonge: Ok, fair enough on the locale, but yes, let's  change the error message

PR #67, opened 2017-10-13
Add deepSearch render variable, Add support for templates

rmberrios:
Add deepSearch render variable, Add support for templates

### DIFF
--- a/docs/controllers.rst
+++ b/docs/controllers.rst
@@ -56,5 +56,14 @@ For the simple pattern of having a controller respond to an specific intent the 
     return { reply: 'LaunchIntent.OpenResponse', to: 'die' };
   });
 
+If you receive a Display.ElementSelected type request, you could use the same approach for intents and state. Voxa receives this type of request and turns it into ``DisplayElementSelected`` Intent
+
+.. code-block:: javascript
+
+  skill.onIntent('DisplayElementSelected', (alexaEvent) => {
+    return { reply: 'DisplayElementSelected.OpenResponse', to: 'die' };
+  });
+
+
 
 Under the hood this creates a new key in the ``entry`` controller and a new state
--- a/docs/views-and-variables.rst
+++ b/docs/views-and-variables.rst
@@ -64,7 +64,7 @@ While the ``tell`` and ``ask`` view types are an exact representation of the bas
 Directives
 -----------
 
-You can include directives in your views file. This will allow you to set directive for Display.RenderTemplate
+Now you can include directives in your views file. Hint directive can be easily specify with the field hint in the views file.
 
 .. code-block:: javascript
 
@@ -75,15 +75,9 @@ You can include directives in your views file. This will allow you to set direct
           'Hello! <break time="3s"/> Good  {time}. Is there anything i can do to help you today?',
           'Hi there! <break time="3s"/> Good  {time}. How may i be of service?',
           'Good  {time}, Welcome!. How can i help you?',
-        ]
+        ],
+        hint: 'hint',
         directives: [
-          {
-            "type": "Hint",
-            "hint": {
-              "type": "PlainText",
-              "text": "Hint"
-            }
-          },
           {
             type: 'Display.RenderTemplate',
             template: {
--- a/docs/views-and-variables.rst
+++ b/docs/views-and-variables.rst
@@ -64,7 +64,7 @@ While the ``tell`` and ``ask`` view types are an exact representation of the bas
 Directives
 -----------
 
-Now you can include directives in your views file. Hint directive can be easily specify with the field hint in the views file.
+Now you can include directives in your views file. Hint directive can be easily specify with a simple object containing a hint key.
 
 .. code-block:: javascript
 
@@ -76,8 +76,11 @@ Now you can include directives in your views file. Hint directive can be easily 
           'Hi there! <break time="3s"/> Good  {time}. How may i be of service?',
           'Good  {time}, Welcome!. How can i help you?',
         ],
-        hint: 'hint',
+
         directives: [
+          {
+            hint: 'hint',
+          },
           {
             type: 'Display.RenderTemplate',
             template: {
--- a/docs/views-and-variables.rst
+++ b/docs/views-and-variables.rst
@@ -61,7 +61,53 @@ say
 While the ``tell`` and ``ask`` view types are an exact representation of the base alexa programming model, the say views are different. They're an abstraction created by voxa to make it simpler to compose your responses over many state transitions. They don't send a respond to alexa but instead make a state transition internally and continue executing your skill code until there's a ``tell`` or ``ask`` response.
 
 
+Directives
+-----------
 
+You can include directives in your views file. This will allow you to set directive for Display.RenderTemplate
+
+.. code-block:: javascript
+
+  const views = {
+    LaunchIntent: {
+      OpenResponse: {
+        tell: [
+          'Hello! <break time="3s"/> Good  {time}. Is there anything i can do to help you today?',
+          'Hi there! <break time="3s"/> Good  {time}. How may i be of service?',
+          'Good  {time}, Welcome!. How can i help you?',
+        ]
+        directives: [
+          {
+            "type": "Hint",
+            "hint": {
+              "type": "PlainText",
+              "text": "Hint"
+            }
+          },
+          {
+            type: 'Display.RenderTemplate',
+            template: {
+              type: "BodyTemplate1",
+              textContent: {
+                primaryText: {
+                    text: "See my favorite car",
+                    type: "PlainText"
+                },
+                secondaryText: {
+                    text: "Custom-painted",
+                    type: "PlainText"
+                },
+                tertiaryText: {
+                    text: "By me!",
+                    type: "PlainText"
+                  }
+              },
+              backButton: "VISIBLE"
+            }
+          }
+      ],
+    },
+  };
 
 Variables
 -----------
@@ -91,6 +137,3 @@ A variable signature is:
         return alexaEvent.locale;
       }
     };
-
-
-
--- a/lib/AlexaSkill.js
+++ b/lib/AlexaSkill.js
@@ -60,6 +60,7 @@ class AlexaSkill {
       'AlexaHouseholdListEvent.ItemsCreated',
       'AlexaHouseholdListEvent.ItemsUpdated',
       'AlexaHouseholdListEvent.ItemsDeleted',
+      'Display.ElementSelected',
     ];
   }
 
--- a/lib/AlexaSkill.js
+++ b/lib/AlexaSkill.js
@@ -131,6 +131,10 @@ class AlexaSkill {
           // call all onRequestStarted callbacks serially.
           return Promise.mapSeries(this.getOnRequestStartedHandlers(), fn => fn(alexaEvent, reply))
             .then(() => {
+              if (alexaEvent.request.type === 'SessionEndedRequest' && _.get(alexaEvent, 'request.reason') === 'ERROR') {
+                throw new Error(_.get(alexaEvent, 'request.error'));
+              }
+
               if (!alexaEvent.session.new) {
                 return null;
               }
--- a/lib/AlexaSkill.js
+++ b/lib/AlexaSkill.js
@@ -10,7 +10,7 @@ const capitalize = require('capitalize');
 const createServer = require('./create-server');
 
 const UnknownRequestType = AlexaError.UnkownRequestType;
-const ErrorOnSessionEnded = AlexaError.ErrorOnSessionEnded;
+const OnSessionEndedError = AlexaError.OnSessionEndedError;
 
 class AlexaSkill {
   constructor(config) {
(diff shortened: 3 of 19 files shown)


lib/renderers/DefaultRenderer.js
```diff
@@ -68,38 +67,65 @@ class DefaultRenderer {
...
-      .value())
-      .then(() => msg);
-  }
+  renderMessage(msg, event) {
+    function deepSearchStatement(statement, alexaEvent) {
+      if (_.isObject(statement) && !_.isArray(statement)) {
+        return Promise.try(() => {
+          const objPromises = _.chain(statement)
+          .toPairs()
+          .map((item) => {
```
armonge: check into the `_.spread` function for this one

lib/renderers/DefaultRenderer.js
```diff
@@ -68,38 +67,65 @@ class DefaultRenderer {
...
-      if (!variable) return Promise.reject(new Error(`No such variable ${token}`));
-      return Promise.try(() => variable(alexaEvent.model, alexaEvent));
-    });
+      if (_.isString(statement)) {
+        return Promise.try(() => this.renderStatement(statement, alexaEvent));
+      }
 
-    const qAll = Promise.all(qVariables);
+      if (_.isArray(statement)) {
+        return Promise.all(statement.map(statementItem => deepSearchStatement.call(this, statementItem, alexaEvent)));
```
armonge: `Promise.map`

lib/renderers/DefaultRenderer.js
```diff
@@ -68,38 +67,65 @@ class DefaultRenderer {
...
-        if (!statement) return null;
-        if (_.isArray(statement)) statement = _.sample(statement);
-        return this.renderStatement(statement, alexaEvent)
-          .then(rendered => _.set(msg, key, rendered));
-      })
-      .compact()
-      .value())
-      .then(() => msg);
-  }
+  renderMessage(msg, event) {
```
armonge: seems like this would require some more tests

test/views.js
```diff
@@ -16,7 +16,28 @@ const views = (function views() {
...
-      Farewell: { tell: 'Ok. For more info visit {site} site.' },
+      Farewell: {
+        tell: 'Ok. For more info visit {site} site.',
+        directives: {
+          type: 'Display.RenderTemplate',
+          template: {
+            type: 'BodyTemplate1',
+            backButton: 'HIDDEN',
+            backgroundImage: {
+              sources: [
```
armonge: will this support creating the array from a variable? if so, we need to test it

`sources: '{variable}'`

docs/controllers.rst
```diff
@@ -56,5 +56,14 @@ For the simple pattern of having a controller respond to an specific intent the
     return { reply: 'LaunchIntent.OpenResponse', to: 'die' };
   });
 
+If you receive a Display.ElementSelected type request, you could use the same approach for intents and state. Voxa receives this type of request and turns it into ``DisplayElementSelected`` Intent
+
+.. code-block:: javascript
+
+  skill.onIntent('DisplayElementSelected', (alexaEvent) => {
```
heneryville: I'm not really comfortable with our shadowing events and other types into intents. It's not really what the user would expect and somewhat breaks the mental-model of Alexa. Can we instead have a differently named handler for this.

docs/views-and-variables.rst
```diff
@@ -61,7 +61,53 @@ say
 While the ``tell`` and ``ask`` view types are an exact representation of the base alexa programming model, the say views are different. They're an abstraction created by voxa to make it simpler to compose your responses over many state transitions. They don't send a respond to alexa but instead make a state transition internally and continue executing your skill code until there's a ``tell`` or ``ask`` response.
 
 
+Directives
+-----------
+
+You can include directives in your views file. This will allow you to set directive for Display.RenderTemplate
```
heneryville: Grammer: This sentence isn't saying anything to me, "This will allow you to set directive for Display.RenderTemplate"

docs/views-and-variables.rst
```diff
@@ -61,7 +61,53 @@ say
...
+
+  const views = {
+    LaunchIntent: {
+      OpenResponse: {
+        tell: [
+          'Hello! <break time="3s"/> Good  {time}. Is there anything i can do to help you today?',
+          'Hi there! <break time="3s"/> Good  {time}. How may i be of service?',
+          'Good  {time}, Welcome!. How can i help you?',
+        ]
+        directives: [
```
heneryville: I wonder if we should try to add any short-hands for these. e.g. just a hint field.

lib/Reply.js
```diff
@@ -141,10 +143,26 @@ class Reply {
...
+
+  if (_.find(directives, { type: 'AudioPlayer.Play' }) && _.find(directives, { type: 'VideoApp.Launch' })) {
+    throw new Error('Do not include both an AudioPlayer.Play directive and a VideoApp.Launch directive in the same response');
+  }
+
+  return directives
+  .filter((directive) => {
+    const hasDisplayInterface = _.get(alexaEvent, 'context.System.device.supportedInterfaces.Display');
+    const directivesType = directive.type;
+    if (!hasDisplayInterface && _.includes(['Display.RenderTemplate', 'Hint'], directivesType)) return false;
```
heneryville: This is key piece of functionality. Can we pull it out into a separate function to make it clearer?

lib/StateMachineSkill.js
```diff
@@ -34,6 +34,15 @@ class StateMachineSkill extends AlexaSkill {
       return this.requestHandlers.IntentRequest(alexaEvent, reply);
     }, true);
 
+    // we treat onDisplay.ElementSelected as an intentRequest
+    this['onDisplay.ElementSelected']((alexaEvent, reply) => {
+      const intent = 'DisplayElementSelected';
+      _.set(alexaEvent, 'intent.name', intent);
+      _.set(alexaEvent, 'intent.slots', {});
+
+      return this.requestHandlers.IntentRequest(alexaEvent, reply);
```
heneryville: Again, I'm worried about making this an intent. If it stays it's own type, we can help with giving a better interface to things like what was clicked, kinda like we provide a better interface to the slots for intents.

lib/renderers/DefaultRenderer.js
```diff
@@ -68,38 +67,72 @@ class DefaultRenderer {
...
+    function deepSearchStatement(statement, alexaEvent) {
```
heneryville: Let's got some clearer comments and start using better named functions to explain what we're doing here. Also unit tests. Also, how much validation should we do on the render cmd object? Before we only rendered certain fields, but I wonder now if we just render everything.

lib/renderers/DefaultRenderer.js
```diff
@@ -68,38 +67,72 @@ class DefaultRenderer {
...
-      const data = _.fromPairs(tokensWithVars);
+      return Promise.resolve(statement);
+    }
 
-      return template(statement, data);
-    });
+    return deepSearchStatement.call(this, msg, event);
+  }
+
+  renderStatement(statement, alexaEvent) {
```
heneryville: Yeah. It's time to start applying software engineering principles to this codebase to make things clearer. We've got too many if's based on data type, and too much special knowlege w/o names.

docs/controllers.rst
rmberrios: Amazon recommends this approach please take a look at this

> There is no built-in intent for selecting actions or list items. However, you can create an intent for this purpose and include it in the intent schema. This intent should be activated when the skill receives a Display.ElementSelected event in a response.

https://developer.amazon.com/docs/custom-skills/display-interface-reference.html#touch-selection-events

docs/views-and-variables.rst
rmberrios: I add a hint field for the views file!

lib/Reply.js
rmberrios: Done!

lib/StateMachineSkill.js
rmberrios: I provide a new way to access this token -> AlexaEvent.token

lib/renderers/DefaultRenderer.js
rmberrios: I create new function with better names

rmberrios:
@heneryville Please check my PR

lib/Reply.js
```diff
@@ -26,9 +26,11 @@ class Reply {
     this.msg = {
       statements: [],
       reprompt: '',
+      hint: undefined,
```
armonge: I'm not sure i like this, why does `hint` gets special treatment instead of just going in the `directives` array?

armonge: i mean, i do like the fact you can declare it like this in the view. What i don't like is the fact that we don't treat it internally as just another directive

lib/Reply.js
```diff
@@ -192,6 +186,16 @@ function cannonicalizeDirective(directive) {
   // Legacy support.
   // We used to support this slightly altered from spec version of the play directive.
   // We'll check if we think that's still happening and transform to the full Alexa directive
+  if (_.isString(directive.hint) && !_.isEmpty(directive.hint)) {
```
armonge: @rmberrios i think this comments are wrong

rmberrios: @armonge  That's related to the play directive. I think we should take it off

lib/AlexaSkill.js
```diff
@@ -131,6 +131,10 @@ class AlexaSkill {
           // call all onRequestStarted callbacks serially.
           return Promise.mapSeries(this.getOnRequestStartedHandlers(), fn => fn(alexaEvent, reply))
             .then(() => {
+              if (alexaEvent.request.type === 'SessionEndedRequest' && _.get(alexaEvent, 'request.reason') === 'ERROR') {
+                throw new Error(_.get(alexaEvent, 'request.error'));
```
armonge: perhaps we should have some custom error so we can check for error type in the error handler? 
https://github.com/mediarain/voxa/blob/master/lib/Errors.js

lib/renderers/DefaultRenderer.js
armonge: @rmberrios what changes did you make regarding this comment?

rmberrios: I didn't apply any change about this comment. We need to define which principles we want to stick to.

armonge: it's related to this other, https://github.com/mediarain/voxa/pull/67#discussion_r145260960 what are the "better names"?

lib/Errors.js
```diff
@@ -32,6 +32,17 @@ class UnkownRequestType extends Error {
   }
 }
 
+class ErrorOnSessionEnded extends Error {
```
armonge: usual naming would be `OnSessionEndedError`

lib/Reply.js
```diff
@@ -140,17 +149,52 @@ class Reply {
 
 }
 
+function hasSupportForDisplay(alexaEvent) {
```
rmberrios: Here is a better name. I applied it to the ones i created

lib/AlexaSkill.js
```diff
@@ -5,10 +5,13 @@ const Reply = require('./Reply');
 const AlexaEvent = require('./AlexaEvent');
 const _ = require('lodash');
 const debug = require('debug')('voxa');
-const UnknownRequestType = require('./Errors').UnkownRequestType;
+const AlexaError = require('./Errors');
 const capitalize = require('capitalize');
 const createServer = require('./create-server');
 
+const UnknownRequestType = AlexaError.UnkownRequestType;
+const OnSessionEndedError = AlexaError.OnSessionEndedError;
```
rmberrios: Rename it to OnSessionEndedError

lib/renderers/DefaultRenderer.js
```diff
@@ -68,7 +68,13 @@ class DefaultRenderer {
   }
 
   renderMessage(msg, event) {
-    function deepSearchStatement(statement, alexaEvent) {
+    /**
+     * It searchs deep looking for string that could have a variable on it
+     * @param  any statement
+     * @param AlexaEvent alexaEvent
+     * @return Promise             Promise with the statment rendered
```
armonge: `statement`

lib/renderers/DefaultRenderer.js
```diff
@@ -68,7 +68,13 @@ class DefaultRenderer {
   }
 
   renderMessage(msg, event) {
-    function deepSearchStatement(statement, alexaEvent) {
+    /**
+     * It searchs deep looking for string that could have a variable on it
+     * @param  any statement
```
armonge: perhaps an example of what a statement looks like and how it would look like after rendering

lib/renderers/DefaultRenderer.js
```diff
@@ -68,7 +68,13 @@ class DefaultRenderer {
   }
 
   renderMessage(msg, event) {
-    function deepSearchStatement(statement, alexaEvent) {
+    /**
+     * It searchs deep looking for string that could have a variable on it
```
armonge: `it makes a deep search for strings...`

armonge:
@heneryville any more comments?

PR #70, opened 2017-10-26
Apiai

rmberrios:


armonge:
Wow, i don't really wanna merge all this work into dev yet. We still need to test some more, have a session to check how we're doing things and see if this is really the direction we want to go

rmberrios:
But why close it? i was working on the PR

PR #74, opened 2017-10-30
Yield on dialog directive

heneryville:
When a dialog directive is sent, the skill should yield. These changes are required to make dialog delegation work in voxa.

### DIFF
--- a/lib/Reply.js
+++ b/lib/Reply.js
@@ -27,7 +27,7 @@ class Reply {
       statements: [],
       reprompt: '',
       card: undefined,
-      yield: false,
+      end: false,
       hasAnAsk: false,
       directives: [],
     };
@@ -44,7 +44,7 @@ class Reply {
 
     const statement = msg.ask || msg.tell || msg.say;
     if (statement) {
-      if (this.msg.yield) {
+      if (this.isYielding()) {
         throw new Error('Can\'t append to already yielding response');
       }
       this.msg.statements.push(statement);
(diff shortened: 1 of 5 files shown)


lib/Reply.js
```diff
@@ -80,12 +80,21 @@ class Reply {
...
+    this.msg.end = true;
     return this;
   }
 
+  hasDirective(type) {
+    return this.msg.directives.some((directive) => {
+      if (_.isRegExp(type)) return !!type.exec(directive.type);
+      if (_.isString(type)) return type === directive.type;
+      if (_.isFunction(type)) return type(directive);
+      throw new Error(`Do not now how to use a ${typeof type} to find a directive`);
```
armonge: @heneryville typo

heneryville:
@armonge I've added several new changes.

lib/Reply.js
```diff
@@ -23,12 +23,13 @@ class Reply {
...
     this.session = alexaEvent.session;
+
     this.msg = {
-      statements: [],
-      reprompt: '',
-      card: undefined,
-      yield: false,
-      hasAnAsk: false,
+      statements: [], // Statements will be concatenated together before being sent
+      reprompt: '', // Since only one reprompt is possible, only the latest value is kept
```
armonge: @heneryville seems to me we should support building reprompts the same way we build ask or tell statements.

armonge:
@heneryville also, please fix or update the failing tests

lib/Reply.js
heneryville: I'm not sure that makes sense. A reprompt should normally be a short single question rather than a long statement that could be constructed. A reprompt with two questions in it is confusing. Heck, if there are ever two reprompts it's probably an error, but for flexibility, overwriting the last is also reasonable.

armonge: My reasoning is that the rendering system should behave similar everywhere, seems to me that while using the system it would feel more natural to expect the reprompt to work the same way as the `say` key

Issue #73, opened 2017-10-30
Unhelpful error messages


heneryville:
When Alexa responds with an error message, voxa doesn't log it very effectively.

 
![screen shot 2017-10-30 at 4 08 15 pm](https://user-images.githubusercontent.com/1660429/32198160-949f387e-bd8c-11e7-9cde-c1528e9d5f5b.png)

heneryville:
Fixed in #75

PR #75, opened 2017-10-30
Stringify error message

wuelcas:
Fix issue #73

### DIFF
--- a/lib/Errors.js
+++ b/lib/Errors.js
@@ -35,9 +35,7 @@ class UnkownRequestType extends Error {
 class OnSessionEndedError extends Error {
   constructor(errorOnSession) {
     if (errorOnSession instanceof Object && errorOnSession.constructor === Object) {
-      const replacer = null;
-      const space = 2;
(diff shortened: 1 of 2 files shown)


lib/Errors.js
```diff
@@ -34,6 +34,11 @@ class UnkownRequestType extends Error {
 
 class OnSessionEndedError extends Error {
   constructor(errorOnSession) {
+    if (errorOnSession instanceof Object && errorOnSession.constructor === Object) {
+      const replacer = null;
+      const space = 2;
+      errorOnSession = JSON.stringify(errorOnSession, replacer, space);
```
omenocal: Do we really need to create a variable for replace and space?
Can we just do `JSON.stringify(errorOnSession, null, 2);`

armonge: Agree with octavio, i understand those variables are helpful if you don't know what the null and 2 values mean, but in this case JSON.stringify is a common enough function with javascript devs that we don't really need it

wuelcas: Ok, I'll delete them then

PR #79, opened 2017-11-04
Validity of slot values

szankelDvag:
Hi I added some handler for the validity of slot values, since often times it happens that you say something else and alexa goes to the intent anyway.

Please check.

Best


szankelDvag:
Will do. All the best.

szankelDvag:
Hi your linter config is out of state. I think it is good to check for validity, cause you can define intent just as slots and then check if the value 100% matches. Beside that you can fill a dynamodb with the values that the user used and which are not in your interaction model to enhance your skill.

Maybe you can fix the lint errors. I can see which they should be.

All the best,

SZankelDvag

armonge:
@szankelDvag you can see the logs for the CI build at https://travis-ci.org/mediarain/voxa/jobs/297175061

Or, you could also run the lint locally with ./run-ci.sh or npm run lint

szankelDvag:
Hi I think you have a couple of more problems. I can't fix them all. Seems like you don't use your eslint conf file in the CI. 

I also added that you can pass back the context to build more convenient state machines. I would also suggest you to add 

"comma-dangle": "off",

to build waiting states where you can react differently on each intent after some question of alexa and of course:

"no-console": "off"

to your eslint conf for cloudwatch.

But other then that your framework is great, keep up the good work.

Sebastian

PR #85, opened 2017-12-01
Add support for playback controller command issued events

rmberrios:
Add support for playback controller command issued events

### DIFF
--- a/lib/StateMachineSkill.js
+++ b/lib/StateMachineSkill.js
@@ -34,14 +34,28 @@ class StateMachineSkill extends AlexaSkill {
       return this.requestHandlers.IntentRequest(alexaEvent, reply);
     }, true);
 
-    // we treat onDisplay.ElementSelected as an intentRequest
-    this['onDisplay.ElementSelected']((alexaEvent, reply) => {
-      const intent = 'DisplayElementSelected';
-      _.set(alexaEvent, 'intent.name', intent);
-      _.set(alexaEvent, 'intent.slots', {});
+    const requestToTurnIntoIntent = [
+      'onDisplay.ElementSelected',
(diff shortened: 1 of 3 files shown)


lib/StateMachineSkill.js
```diff
@@ -34,14 +34,28 @@ class StateMachineSkill extends AlexaSkill {
       return this.requestHandlers.IntentRequest(alexaEvent, reply);
```
armonge: why have this as a separate piece? maybe we should also put the LaunchRequest handling with the rest of `requestToTurnIntoIntent`

rmberrios: Done!

lib/StateMachineSkill.js
```diff
@@ -43,12 +35,18 @@ class StateMachineSkill extends AlexaSkill {
...
-          .drop(2)
-          .join('')
-          .replace('.', '')
-          .value();
+        let intent;
+
+        if (requestToConvert === 'onLaunchRequest') {
+          intent = this.config.openIntent || 'LaunchIntent';
+        } else {
+          intent = _.chain(requestToConvert)
```
armonge: this is also something hard to ready, maybe we could split it into a separate function? Like, `requestNameToIntent` or something like that is easier to understand than
```js
            .split('')
            .drop(2)
            .join('')
            .replace('.', '')
            .value();
```

We could also have a test that shows `onPlaybackController.PlayCommandIssued` is transformed into `PlaybackControllerPlayCommandIssuedIntent` will make us understand why the need for all those 'replace', drop, etc

rmberrios: Done!

PR #125, opened 2018-07-05
Added CanFulfillIntentRequest feature

omenocal:
For more information about `CanFulfillIntentRequest`, visit these 2 links:

https://developer.amazon.com/docs/custom-skills/understand-name-free-interaction-for-custom-skills.html
https://developer.amazon.com/docs/custom-skills/quick-start-canfulfill-intent-request.html

The way it's meant to work is that in every intent should validate the scenario when this value is coming:

```
skill.onIntent('MyIntent', (alexaEvent) => {
    if (alexaEvent.request.isCanFulfillIntentRequest) {
        const canFulfillIntent = { canFulfill: 'YES' };
        return { canFulfillIntent };
    }

    // DO YOUR NORMAL CODE
});
```

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -12,7 +12,7 @@ Voxa is a framework that provides a way to organize a conversational experience 
 
 Why Voxa vs other frameworks
 ----------------------------
-Voxa provides a more robust framework for building Alexa skills, Cortana skills, or Google Assistant Actions. It provides a design pattern that wasn’t found in other frameworks. Critical to Voxa was providing a pluggable interface and supporting all of the latest platform features.
+Voxa provides a more robust framework for building Alexa skills. It provides a design pattern that wasn't found in other frameworks. Critical to Voxa was providing a pluggable interface and supporting all of the latest ASK features.
 
 Features
 --------
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -8,11 +8,11 @@ Welcome to Voxa's documentation!
 
 Summary
 -------
-Voxa is an Alexa skill framework that provides a way to organize a skill into a state machine. Even the most complex voice user interface (VUI) can be represented through the state machine and it provides the flexibility needed to both be rigid when needed in specific states and flexible to jump around when allowing that also makes sense.  
+Voxa is a framework that provides a way to organize a conversational experience into a state machine. Even the most complex voice user interface (VUI) can be represented through the state machine and it provides the flexibility needed to both be rigid when needed in specific states and flexible to jump around states when needed.
 
 Why Voxa vs other frameworks
 ----------------------------
-Voxa provides a more robust framework for building Alexa skills.  It provides a design pattern that wasn’t found in other frameworks.   Critical to Voxa was providing a pluggable interface and supporting all of the latest ASK features.  
+Voxa provides a more robust framework for building Alexa skills, Cortana skills, or Google Assistant Actions. It provides a design pattern that wasn’t found in other frameworks. Critical to Voxa was providing a pluggable interface and supporting all of the latest platform features.
 
 Features
 --------
--- a/docs/reply.rst
+++ b/docs/reply.rst
@@ -12,12 +12,12 @@ The ``reply`` Object
   :param message: A message object
 
   .. js:function:: Reply.append(message)
-    
+
     Adds statements to the ``Reply``
-    
-    :param message: An object with keys ``ask``, ``tell``, ``say``, ``reprompt``, ``card`` or ``directives`` keys. Or another ``reply`` object
+
+    :param message: An object with keys ``ask``, ``tell``, ``say``, ``reprompt``, ``card``, ``canFulfillIntent`` or ``directives`` keys. Or another ``reply`` object
     :returns: the ``Reply`` object
 
   .. js:function:: Reply.toJSON()
-    
+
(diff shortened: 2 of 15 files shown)


docs/index.rst
```diff
@@ -8,11 +8,11 @@ Welcome to Voxa's documentation!
...
+Voxa is a framework that provides a way to organize a conversational experience into a state machine. Even the most complex voice user interface (VUI) can be represented through the state machine and it provides the flexibility needed to both be rigid when needed in specific states and flexible to jump around states when needed.
 
 Why Voxa vs other frameworks
 ----------------------------
-Voxa provides a more robust framework for building Alexa skills.  It provides a design pattern that wasn’t found in other frameworks.   Critical to Voxa was providing a pluggable interface and supporting all of the latest ASK features.  
+Voxa provides a more robust framework for building Alexa skills, Cortana skills, or Google Assistant Actions. It provides a design pattern that wasn’t found in other frameworks. Critical to Voxa was providing a pluggable interface and supporting all of the latest platform features.
```
armonge: Voxa 2 has no support for Cortana or Google assistant. Let's make this changes on the v3

lib/AlexaSkill.js
```diff
@@ -114,11 +115,15 @@ class AlexaSkill {
         }
       }
 
-
       if (this.getOnLaunchRequestHandlers().length === 0) {
         throw new Error('onLaunchRequest must be implemented');
       }
 
+      if (alexaEvent.request.type === 'CanFulfillIntentRequest') {
```
armonge: Why the need to handle this as an intent?

samples/accountLinking/skill/states.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+const _ = require('lodash');
+
+exports.register = function register(skill) {
```
armonge: is this part of the PR?

samples/accountLinking/skill/states.js
```diff
@@ -0,0 +1,29 @@
...
+    if (_.get(alexaEvent, 'model.user.email')) {
+      return { reply: 'Intent.Launch', to: 'entry' };
+    }
+
+    return { reply: 'Intent.NotAuthenticated', to: 'die' };
+  });
+
+  skill.onIntent('AMAZON.HelpIntent', () => ({ reply: 'Intent.Help' }));
+
+  skill['onAlexaSkillEvent.SkillEnabled']((alexaEvent, reply) => {
```
armonge: but it does seem like we should be able to handle this event the same way we handle this other custom events

samples/accountLinking/skill/states.js
omenocal: This looked like it was merged directly to master instead of DEV, so now that I try to sync branches, I had to include it as part of the conflicts

samples/accountLinking/skill/states.js
omenocal: I think I got your point, but this is how we are doing it in other skills:

https://github.com/mediarain/mars-skill/blob/staging/skill/events.js#L14
https://github.com/mediarain/vaseline-skill/blob/staging/skill/states/events-state.js#L25

lib/AlexaSkill.js
omenocal: Every intent can be fulfilled. To verify that, every intent with their slots should have the logic if they need to be fulfilled or not. That's why I convert this request to an `IntentRequest` and add this boolean flag so developers know if they're going to handle this canFulfillRequest. If they deny, the flow ends and the state machine sends the response back to the Alexa service. If they do accept to fulfill it, the state machine starts over the process with this flag set to `undefined`

rmberrios:
After discussing with @omenocal we decided to go this way:
```
skill.onCanFulfillIntentRequest((alexaEvent) => {});
```

lib/AlexaSkill.js
heneryville: I'm not sure that I love that approach. For example, not all intents are going to be on the root level. Also, isn't it odd the co-mingle the logic for how to kick off an intent, and report that it is actionable in one handler?

It seems like this is putting too much work into intent handlers, and with special flags to indicate something is slightly different. Also, there's likely to be only a few intents that you'd want to signal as can fulfill, but the rest should be no by default. This way all of the intents are going to need to do this special check.

Finally, it seems like we should offer a shorthand expression for intents that are fulfillable regardless of slot, then a way to say, let me run a function to be sure.

heneryville:
Agreed about avoiding imitating this as an intent. I know we've done that elsewhere, but it was a mistake. We should be more faithful to the how Alexa actually works.

docs/reply.rst
```diff
@@ -12,12 +12,12 @@ The ``reply`` Object
   :param message: A message object
 
   .. js:function:: Reply.append(message)
-    
+
     Adds statements to the ``Reply``
-    
-    :param message: An object with keys ``ask``, ``tell``, ``say``, ``reprompt``, ``card`` or ``directives`` keys. Or another ``reply`` object
+
+    :param message: An object with keys ``ask``, ``tell``, ``say``, ``reprompt``, ``card``, ``canFulfillIntent`` or ``directives`` keys. Or another ``reply`` object
```
armonge: @omenocal seems like we have to update this to reflect the latest refactor?

PR #132, opened 2018-07-20
Add intent when there is an intent for Alexa

heneryville:
This odd line of code was preventing `AlexaEvent` from ever making it's `AlexaIntent`.


heneryville:
It turns out that this is a bad way to approach this. That check is trying to guarantee that an intent property is always available. I'm not sure why that's a good idea, but there appears to be things that rely on that.

PR #133, opened 2018-07-25
Added CanFulfillIntentRequest feature

omenocal:


docs/canFulfillIntentRequest.rst
```diff
@@ -0,0 +1,56 @@
...
+Additionally, if you have several intents that you want to automatically fulfill, regardless of the slot values in the request, you can simply add an array of intents to the property: `defaultFulfillIntents`_ of the Voxa config file:
+
+.. code-block:: javascript
+
+  const defaulFulfillIntents = [
+    'NameIntent',
+    'PhoneIntent',
+  ];
+
+  const skill = new StateMachineSkill({ variables, views, defaultFulfillIntents });
```
armonge: this seems like voxa2

docs/reply.rst
```diff
@@ -21,3 +21,13 @@ The ``reply`` Object
   .. js:function:: Reply.toJSON()
```
armonge: let's take the oportunity to update the Reply object docs, we don't have a toJSON for example

docs/reply.rst
```diff
@@ -21,3 +21,13 @@ The ``reply`` Object
...
 
     :returns: An object with the proper format to send back to Alexa, with statements wrapped in SSML tags, cards, reprompts and directives
+
+  .. js:function:: Reply.fulfillIntent(canFulfill)
```
armonge: and this is specific to the AlexaReply object

src/VoxaApp.ts
```diff
@@ -16,6 +16,7 @@ const log: debug.IDebugger = debug("voxa");
 
 export interface IVoxaAppConfig extends IRendererConfig {
   appIds?: string[]|string;
+  defaultFulfillIntents?: string[];
```
armonge: as this is platform specific, let's move it to the AlexaPlatform constructor,  something like this https://github.com/mediarain/voxa/blob/v3/src/platforms/botframework/BotFrameworkPlatform.ts#L50

test/VoxaApp.spec.ts
```diff
@@ -312,6 +313,151 @@ describe("VoxaApp", () => {
...
+
+    const voxaApp = new VoxaApp({ views, variables });
+    const alexaSkill = new AlexaPlatform(voxaApp);
+    voxaApp.onCanFulfillIntentRequest((alexaEvent: AlexaEvent, alexaReply: AlexaReply) => {
+      alexaReply.fulfillIntent("YES");
+      alexaReply.fulfillSlot("slot1", "YES", "YES");
+      return alexaReply;
+    });
+
+    const fulfillEvent = _.cloneDeep(event);
```
armonge: we have a AlexaRequestBuilder to be used in tests, please update it to support this request type and let's use that instead

https://github.com/mediarain/voxa/blob/v3/test/tools.ts#L16

docs/transition.rst
```diff
@@ -39,7 +39,7 @@ Directives are used passed directly to the alexa response, the format is describ
 ``reply``
 ---------
 
-The ``reply`` key can take 2 forms, a simple string pointing to one of your views or a :ref:`Reply <reply>` object.
+The ``reply`` key can take 2 forms, a simple string pointing to one of your views or a :ref:`Reply <voxa-reply>` object.
```
armonge: we don't have a `reply` key anymore, thinking of adding it back

PR #135, opened 2018-07-31
Fixed gadget animation builder issue

omenocal:


armonge:
so, what was the issue?

omenocal:
@armonge it was adding an extra object every time the `setAnimations ` method was called

armonge:
In the future let's try and add those comments to PRs

PR #139, opened 2018-08-10
Voxa SSML appending works with SSML that contains new lines

heneryville:
There was a bug in the SSML append code that would fail to unwrap the `<speak>` tags from SSML that contained newlines. This is a fix.

I also took an opportunity to re-structure some code in the StateMachine to make it a touch more self-documenting.


armonge:
retest retry

armonge:
@heneryville let's fix the tests

heneryville:
@armonge Got that cleaned up. How is this now?

PR #140, opened 2018-08-13
Data access changes

heneryville:
This change is a proposal to slightly modify how we do serialization/deserialization.
Previously we'd stick everything into the attributes in the same object level as the model. This would lead to unexpected behavior if the model was cleared or the risk of model data and plugin data clobbering one another.

This change introduces two paths to data serialization/deserialization.

1) Models. Models continue to use there serialize method. `fromEvent` has been renamed `deserialize` and is now *passed* the model data, rather than making the implementor dig it out of session.attributes itself. Besides making the interface more obvious, this could allow us to in the future inject other middleware that knows how to retrieve additional model data (e.g. from a database) w/o the models needing the change. Deserialize now goes into the reply attributes under the `model` key.

2) The session now has an attribute called `outputAttributes`. This is where non-model things put their attributes (e.g. plugins, or where the state goes). It's now a bit clearer that this is separate from the model. Of course plugins can still clobber each other, but at least the model, the most common thing for developers to access, is isolated now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "prepare": "npm run clean && tsc",
     "compile": "tsc",
     "lint": "tslint --project tsconfig.json --config tslint.json ",
+    "lint-fix": "tslint --project tsconfig.json --config tslint.json --fix ",
     "report": "nyc report --reporter=json && nyc report --reporter html && nyc report --reporter=lcov && nyc report --reporter=cobertura",
     "test": "mocha test/*.spec.* test/**/*.spec.*",
     "test-ci": "nyc mocha --colors --reporter mocha-jenkins-reporter test test/*.spec.* test/**/*.spec.*"
--- a/package.json
+++ b/package.json
@@ -9,7 +9,6 @@
     "prepare": "npm run clean && tsc",
     "compile": "tsc",
     "lint": "tslint --project tsconfig.json --config tslint.json ",
-    "lint-fix": "tslint --project tsconfig.json --config tslint.json --fix ",
     "report": "nyc report --reporter=json && nyc report --reporter html && nyc report --reporter=lcov && nyc report --reporter=cobertura",
     "test": "mocha test/*.spec.* test/**/*.spec.*",
     "test-ci": "nyc mocha --colors --reporter mocha-jenkins-reporter test test/*.spec.* test/**/*.spec.*"
--- a/src/VoxaApp.ts
+++ b/src/VoxaApp.ts
@@ -102,8 +102,8 @@ export class VoxaApp {
   }
 
   public validateConfig() {
-    if (!this.config.Model.fromEvent) {
-      throw new Error("Model should have a fromEvent method");
+    if (!this.config.Model.deserialize) {
+      throw new Error("Model should have a deserialize method");
     }
 
     if (!this.config.Model.serialize && !(this.config.Model.prototype && this.config.Model.prototype.serialize)) {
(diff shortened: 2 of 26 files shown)


package.json
```diff
@@ -9,6 +9,7 @@
     "prepare": "npm run clean && tsc",
     "compile": "tsc",
     "lint": "tslint --project tsconfig.json --config tslint.json ",
+    "lint-fix": "tslint --project tsconfig.json --config tslint.json --fix ",
```
armonge: no real need for this since latest npm and yarn let you add paremeters like

`yarn lint --fix`
`npm run lint -- --fix`

src/StateMachine.ts
```diff
@@ -55,6 +56,29 @@ export function isState(object: any): object is IState {
   return object  && "name" in object ;
 }
 
+// A helper class for transitions. Users can return transitions as an object with various command keys.
```
armonge: wan't this already in the other PR?

src/VoxaApp.ts
```diff
@@ -468,23 +468,29 @@ export class VoxaApp {
...
-      voxaEvent.model.state = transition.to;
-    } else if (isState(transition.to)) {
-      voxaEvent.model.state = transition.to.name;
-    }
-
-    await response.saveSession(voxaEvent);
+    const stateName = typeof transition.to === "string" ? transition.to
+    : isState(transition.to) ? transition.to.name
+    : "";
+    if (!stateName) { throw new Error("Expected transition to transition to something"); }
```
armonge: let's add a custom error type to https://github.com/mediarain/voxa/tree/v3/src/errors

src/platforms/botframework/BotFrameworkReply.ts
```diff
@@ -205,7 +205,7 @@ export class BotFrameworkReply implements IVoxaReply {
     return await rp(requestOptions) as IAuthorizationResponse;
   }
 
-  public async saveSession(event: IVoxaEvent): Promise<void> {
+  public async saveSession(attributes: object, event: IVoxaEvent): Promise<void> {
```
armonge: this should probably be `IBag`

test/create-server.spec.ts
```diff
@@ -63,6 +63,7 @@ describe("createServer", () => {
         },
         shouldEndSession: true,
       },
+      sessionAttributes: {},
```
armonge: i'd rather we don't include `sessionAttributes` in the response if it's empty

package.json
heneryville: Did not know. Cool!

src/VoxaApp.ts
heneryville: Added

test/create-server.spec.ts
heneryville: I agree, and tried to do it, but it was rather hard to make consistent. Basically the to do it I'd need to make the sessionAttributes undefined conditionally. I could do that in saveSession, but not all of the tests actually call saveSession. After messing with it for a while, I just decided that consistently there was better than sometimes there and empty, and sometimes not.

package.json
heneryville: It looks like yarn supports this, but NPM does not.

heneryville:
Ok. All issues addressed. This is ready for another look.

test/Gadgets.spec.ts
```diff
@@ -208,7 +208,6 @@ describe("Gadgets", () => {
     expect(_.get(reply, "response.outputSpeech.ssml")).to.include("Thanks for playing with echo buttons.");
     expect(reply.response.reprompt).to.be.undefined;
     expect(_.get(reply, "response.directives[0].type")).to.equal("GameEngine.StopInputHandler");
-    expect(_.get(reply, "sessionAttributes.originatingRequestId")).to.equal("originatingRequestId");
```
armonge: @omenocal could you check this?

omenocal: Yes, I'm fine with removing this line, I will move it onto another test in my next PR.

PR #146, opened 2018-08-28
Added Device Settings API

omenocal:
For more info, visit the public documentation:
https://developer.amazon.com/docs/smapi/alexa-settings-api-reference.html

For some reason the temperature endpoint is not working. Waiting for AMZ to send any fix and add it to Voxa (if needed)


lib/apis/DeviceSettings.js
```diff
@@ -0,0 +1,92 @@
+'use strict';
+
+const _ = require('lodash');
+const debug = require('debug')('voxa');
+const rp = require('request-promise');
+
+class DeviceSettings {
+  constructor(alexaEvent) {
```
armonge: So, most of this API Classes have a similar format, with a constructor that takes the alexaEvevnt and grabs the apiAccessToken and apiEndpoint, they also have a `getOptions` method, let's go ahead and make an actual base class and share some code

lib/apis/DeviceSettings.js
```diff
@@ -0,0 +1,92 @@
...
+      timezone,
+    };
+  }
+}
+
+/*
+ * Checks if it's safe to ignore errors
+ * https://developer.amazon.com/docs/smapi/alexa-settings-api-reference.html#response-codes
+ */
+function checkError(err) {
```
armonge: this also seems like something we can reuse

PR #149, opened 2018-09-03
Added Alexa Contact, Device, ISP and List APIs

omenocal:
Features added:

- Customer Contact API
- Device Address API
- Device Settings API
- In-Skill Purchases API
- Lists API


docs/index.rst
```diff
@@ -133,6 +133,11 @@ Links
   statemachine-skill
   request-flow
   i18n
+  customerContact
```
armonge: we need a way to signal the user from the menu that this is alexa only

package.json
```diff
@@ -61,6 +61,7 @@
     "@types/chai": "^4.0.10",
     "@types/chai-as-promised": "^7.1.0",
     "@types/mocha": "^2.2.45",
+    "@types/nock": "^9.3.0",
```
armonge: i believe you also need to add `nock`, not only `@types/nock`

src/VoxaEvent.ts
```diff
@@ -24,6 +24,8 @@ export abstract class IVoxaEvent {
   public requestToIntent: ITypeMap = {};
   public requestToRequest: ITypeMap = {};
   public platform!: string;
+  public alexa: any;
```
armonge: actually, this seems like keys that should only be present in the `AlexaEvent` and `DialogFlowEvent` subclasses

armonge: Also, they're not `any` we know which keys are present, go ahead and define a type

docs/inSkillPurchases.rst
```diff
@@ -0,0 +1,82 @@
...
+
+Buying these products in a skill is seamless to a user. They may ask to shop products, buy products by name, or agree to purchase suggestions you make while they interact with a skill. Customers pay for products using the payment options associated with their Amazon account.
+
+For more information about setting up ISP with the ASK CLI follow this `link <https://developer.amazon.com/docs/in-skill-purchase/use-the-cli-to-manage-in-skill-products.html>`_. And to understand what's the process behind the ISP requests and responses to the Alexa Service click `here <https://developer.amazon.com/docs/in-skill-purchase/add-isps-to-a-skill.html>`_.
+
+With Voxa, you can implement all ISP features like buying, refunding and upselling an item:
+
+.. code-block:: javascript
+
+  skill.onIntent('BuyIntent', async (voxaEvent) => {
```
lvelasquezm: I know ISP is an Alexa feature, but shouldn't this be `app` instead of `skill`?

lvelasquezm: Same for all other `skill` statements

src/platforms/alexa/AlexaEvent.ts
```diff
@@ -50,6 +57,14 @@ export class AlexaEvent extends IVoxaEvent {
     }
 
     this.platform = "alexa";
+
+    this.alexa = {
+      customerContact: new CustomerContact(event),
```
armonge: instead of `event` better pass the `rawEvent`

src/platforms/alexa/apis/ApiBase.ts
```diff
@@ -0,0 +1,49 @@
...
+import * as debug from "debug";
+import * as _ from "lodash";
+import * as rp from "request-promise";
+
+const alexalog: debug.IDebugger = debug("voxa:alexa");
+
+export class ApiBase {
+  public errorCodeSafeToIgnore: number = 0; // the code error to ignore on checkError function
+  public tag: string = ""; // the class reference for error logging
+  public voxaEvent: any; // the event as sent by the service
```
armonge: instead of using a `voxaEvent` perhaps it's better to use the `rawEvent` which has a type of RequestEnvelope if i'm not mistaken. In any case, let's try and not use `any` when possible

src/platforms/alexa/apis/ApiBase.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  constructor(event: RequestEnvelope) {
+    this.voxaEvent = _.cloneDeep(event);
+  }
+
+  protected getResult(path = "", method = "GET", body = {}): any {
```
armonge: the type of each api response can be found in the ask-sdk-model package, for example

services.deviceAddress.Address
services.deviceAddress.ShortAddress

Let's use those, seems like a great oportunitty for [generics](https://www.typescriptlang.org/docs/handbook/generics.html)

docs/inSkillPurchases.rst
omenocal: Yeah, I think it makes sense. I was just following the same structure of the docs. If you look at other docs' examples, it uses `skill` as well. Also, in the headspace action and skill, the guys are still using `skill` for their intents/states:
https://github.com/mediarain/headspace-skill/blob/V3-staging/app/states/alexa.states.js

@armonge thoughts on this change ?? If we go with @lvelasquezm suggestion, I'd have to change all the docs (and probably the tests?)

I also think is a matter of the developer's preference. I really like `skill` since I'm coming from the voxa2 version and I'm used to that. But open to any guidance on this topic.

src/platforms/alexa/apis/InSkillPurchase.ts
```diff
@@ -0,0 +1,96 @@
...
+      upsellMessage,
+    };
+
+    return new ConnectionsSendRequest("Upsell", payload, token);
+  }
+
+  public voxaEvent: any; // the event as sent by the service
+
+  constructor(event: RequestEnvelope) {
+    this.voxaEvent = _.cloneDeep(event);
```
armonge: check this, you're passing a `voxaEvent` as any in the calling `AlexaEvent` class, however this class takes a `RequestEnvelope` in the constructor

package.json
omenocal: Hmm I didn't have any issues in my tests nor lint, nor in Travis, do you still think is worth it to include it ?

src/VoxaEvent.ts
omenocal: Ok, I'll remove the google one because I'm not using it. I'll just move out the `alexa` key

docs/inSkillPurchases.rst
armonge: We should call them `app`, in Voxa3 most of the code is part of the VoxaApp, the `skill` is just the actual wrapper. Also, if you see the docs calling stuff `skill` instead of `app` go ahead and change that, wherever it makes sens

package.json
armonge: yes, most likely `nock` was included as a dependency from another package, however the `@types/nock` package does not include it as a dependency. Add `nock` in here to have a specific version and not depend on side effects from other packages

omenocal:
screen shot 2018-09-04 at 11 57 54 am

@armonge done, just added bullets to group Alexa APIs in a single item

PR #157, opened 2018-10-09
Adding a new IVoxaIntentEvent interface

armonge:


src/VoxaApp.ts
```diff
@@ -221,11 +221,13 @@ export class VoxaApp {
               },
             );
 
-            // call all onSessionStarted callbacks serially.
-            await bluebird.mapSeries(
-              this.getOnSessionStartedHandlers(voxaEvent.platform.name),
-              (fn: IEventHandler) => fn(voxaEvent, reply),
-            );
+            if (voxaEvent.session.new) {
```
armonge: this should only run on new sessions, not sure why it was not working like that

src/platforms/alexa/AlexaEvent.ts
```diff
@@ -46,25 +52,10 @@ export class AlexaEvent extends IVoxaEvent {
   };
 
   public requestToIntent: any = {
-    "AlexaSkillEvent.SkillDisabled": "AlexaSkillEvent.SkillDisabled",
```
armonge: all this requests were being interpretted as intents... that's not correct, we don't need to execute the whole state machine for this

PR #158, opened 2018-10-11
State Machine refactor

armonge:
The code for the State Machine was getting too hairy, i refactored it to get a simpler model, create a new State Class, removed some of the features we rarely use and started using an actual `State` class

Basically we no longer have the `entry` special state, that also means we now only have 2 ways of specifying a state, a function handler and a literal transition object. Before there was also the `entry` way, which by the way could also take a function, this was leadomg to harder code in the StateMachine and from looking at our past projects wasn't a feature used often in recent projects


hello-world/hello-world.js
```diff
@@ -51,15 +50,21 @@ app.onBeforeReplySent(request => {
...
-  if (request.intent.name === "NoIntent") {
-    return { ask: "doesNotLikeVoxa" };
-  }
-});
+app.onState(
+  "likesVoxa?",
+  {
+    tell: "doesNotLikeVoxa"
+  },
+  "NoIntent"
```
armonge: This was a feature intended since the beginning, finally got it working with code i actually feel good about

test/VoxaPlatform.spec.ts
```diff
@@ -182,12 +183,14 @@ describe("VoxaPlatform", () => {
   describe("onState", () => {
     let alexaLaunch: any;
     let dialogFlowLaunch: any;
+    let processData: any;
 
     afterEach(() => {
-      delete process.env.DEBUG;
+      process.env = processData;
```
armonge: this was deleting my logs in the tests when running them with `DEBUG=voxa yarn test --watch --bail`

PR #165, opened 2018-10-30
Added LoginWithAmazon to APIs

omenocal:
Amazon docs here:
https://developer.amazon.com/docs/login-with-amazon/customer-profile.html

### DIFF
--- a/lib/apis/LWA.js
+++ b/lib/apis/LWA.js
@@ -1,0 +1,27 @@
+'use strict';
+
+const ApiBase = require('./ApiBase');
+
+class LWA extends ApiBase {
+
+  // eslint-disable-next-line class-methods-use-this
+  getToken() {
+    return undefined;
+  }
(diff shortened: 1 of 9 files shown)


lib/apis/LWA.js
```diff
@@ -0,0 +1,27 @@
...
+  getEndpoint() {
+    return 'https://api.amazon.com';
+  }
+
+  /*
+   * Gets the user's profile information.
+   * https://developer.amazon.com/docs/login-with-amazon/customer-profile.html
+   */
+  getUserInformation() {
+    const accessToken = this.alexaEvent.user.accessToken;
```
armonge: shouldn't this come from `getToken`?

omenocal: `getToken` provides a token for the Authorization headers. LWA adds it as a GET parameter :/

PR #166, opened 2018-11-05
Fixes a bug that prevented the SM from using the global intent handlers

armonge:
The current behavior only allowed for global intent handlers to be called when the intent event came from a new session or the current state was entry. This was a regression that meant the following code wouldn't call the `HelpIntent` handler in the following example when in the `helpSettings` state.

This PR fixes that

```ts
    voxaApp.onIntent("HelpIntent", {
      flow: "yield",
      sayp: "Help",
      to: "entry",
    });

    voxaApp.onState(
      "helpSettings",
      {
        flow: "yield",
        sayp: "question",
        to: "entry",
      },
      "YesIntent",
    );

    voxaApp.onState(
      "helpSettings",
      (voxaEvent: IVoxaEvent) => {
        return {
          flow: "yield",
          sayp: "question",
          to: "entry",
        };
      },
      "NoIntent",
    );
```

### DIFF
--- a/src/StateMachine/StateMachine.ts
+++ b/src/StateMachine/StateMachine.ts
@@ -77,7 +77,12 @@ export class StateMachine {
       throw new Error("State Machine Recursion Error");
     }
 
-    const transition = await this.stateTransition(fromState, voxaEvent, reply);
+    const transition = await this.stateTransition(
+      fromState,
+      voxaEvent,
+      reply,
+      recursions,
+    );
 
     let sysTransition = await this.checkOnUnhandledState(
       voxaEvent,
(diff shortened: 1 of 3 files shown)


src/StateMachine/StateMachine.ts
```diff
@@ -127,13 +133,24 @@ export class StateMachine {
         );
       }
     } catch (error) {
+      /*
+       * Returning to the global handler here only makes sense if we didn't already made that,
+       * meaning that it could only be done in the first recussion. There's tests covering this scenarion
```
omenocal: Typo: scenarion

src/StateMachine/StateMachine.ts
```diff
@@ -127,13 +133,24 @@ export class StateMachine {
...
+       * in tests/States.spec.ts
+       */
+      if (fromState !== "entry" && recursions < 1) {
```
omenocal: In which part of the code is this variable updated ?

armonge: https://github.com/mediarain/voxa/pull/166/files#diff-50197b557be628917c89e4d26537925aR108

PR #167, opened 2018-11-06
Update ask sdk model

wuelcas:
Please review the code first, this code has room for improvement.

The `ask-sdk-model` package has the new types for APL related requests and intents but updating the package lead to several changes for type definitions.

The `set` and `get` methods from `lodash` were mainly used to get a property that is not in all request definitions anymore.

The CanFullfillIntent definition was removed from the `ask-sdk-model` package so the definitions from the beta version we used were added.

### DIFF
--- a/src/platforms/alexa/AlexaCanFullfillReply.ts
+++ b/src/platforms/alexa/AlexaCanFullfillReply.ts
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) 2018 Rain Agency <contact@rain.agency>
+ * Author: Rain Agency <contact@rain.agency>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { ResponseEnvelope } from "ask-sdk-model";
+import * as _ from "lodash";
+import { IBag, IVoxaEvent } from "../../VoxaEvent";
+import { IVoxaReply } from "../../VoxaReply";
+
+export class AlexaCanFullfillReply implements IVoxaReply, ResponseEnvelope {
+  public version = "1.0";
+  public response: any = {};
+  public sessionAttributes: IBag = {};
+
+  get hasMessages() {
+    return false;
+  }
+
+  get hasDirectives() {
+    return false;
+  }
+
+  get hasTerminated() {
+    return true;
+  }
+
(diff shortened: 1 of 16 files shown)


armonge:
@wuelcas we're using prettier and yarn for this project, please update the yarn.lock and fix the errors reported by travis with prettier

armonge:
@wuelcas any updates?

src/platforms/alexa/AlexaPlatform.ts
```diff
@@ -137,7 +137,7 @@ export class AlexaPlatform extends VoxaPlatform {
   }
 
   protected getReply(event: IVoxaEvent): IVoxaReply {
-    return new AlexaReply();
+    return new AlexaReply(event);
```
armonge: i was just thinking, what about instead of having a lot of checks for the type inside the AlexaReply, what about actually having a separate AlexaCanfulfillReply? something that implements the IVoxaReply interface but it's an actually different class

wuelcas: I agree, I'll work on implement that

wuelcas: @armonge Please check out the latest changes

src/platforms/alexa/AlexaCanFullfillReply.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  public get speech(): string {
+    return _.get(this.response, "outputSpeech.ssml", "");
+  }
+
+  public get reprompt(): string {
+    return _.get(this.response, "reprompt.outputSpeech.ssml", "");
+  }
+
+  public addStatement(statement: string, isPlain: boolean = false) {
```
armonge: i think it would be best to throw an exception in all this methods, something like "You tried to add an statement to a canfullfil reply" instead of just not doing anything

src/platforms/alexa/AlexaCanFullfillReply.ts
```diff
@@ -0,0 +1,111 @@
...
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { ResponseEnvelope } from "ask-sdk-model";
+import * as _ from "lodash";
+import { IBag, IVoxaEvent } from "../../VoxaEvent";
+import { IVoxaReply } from "../../VoxaReply";
+
+export class AlexaCanFullfillReply implements IVoxaReply, ResponseEnvelope {
```
armonge: does it really have to implement ResponseEnvelope?

src/platforms/alexa/AlexaCanFullfillReply.ts
```diff
@@ -0,0 +1,111 @@
...
+  public addStatement(statement: string, isPlain: boolean = false) {
+    return;
+  }
+
+  public addReprompt(statement: string, isPlain: boolean = false) {
+    return;
+  }
+
+  public fulfillIntent(canFulfill: any) {
+    _.set(this.response, "card", undefined);
```
armonge: i don't think you need to remove this "card", "reprompt", etc, remember we never initialize them on this replies

test/Renderer.spec.ts
```diff
@@ -112,7 +112,7 @@ describe("Renderer", () => {
           say: "Say",
           to: "entry",
         }));
-        rawEvent.request.locale = locale;
+        _.set(rawEvent.request, "locale", locale);
```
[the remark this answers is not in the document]
wuelcas: Because `locale` doesn't exists in all request types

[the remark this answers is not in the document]
armonge: yes but, those request types usually don't trigger the renderer, also, i don't think IVoxaEvent.request.locale is optional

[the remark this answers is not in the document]
wuelcas: So should I set locale as optional? or maybe pass a default locale? The request with no locale is `SkillEnabledRequest`

src/platforms/alexa/CanFullfillintent.ts
```diff
@@ -0,0 +1,60 @@
+import { DialogState, Intent } from "ask-sdk-model";
```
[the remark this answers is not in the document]
wuelcas: Since we didn't make any modification to the code I think we need to license this particular file as Apache, at least that's what I understand from a quick Google search. So instead of setting the MIT license at the top, we put the Apache license.

test/Renderer.spec.ts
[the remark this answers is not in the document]
armonge: the renderer requires a locale, there's no way around that. What i'm not sure is why there being a request with no locale matters to the renderer code. There's probably something we should be doing so those 2 pieces of code are not as dependant on each other

wuelcas:
I'm closing this PR because the last version of the ask-sdk-model library already have the CanFullfill Request types. I'll be making a new one, much smaller.

PR #168, opened 2018-11-14
Added LoginWithAmazon and Google Sign-in methods

omenocal:
Amazon docs here:
https://developer.amazon.com/docs/login-with-amazon/customer-profile.html

Google Assistant docs here:
https://developers.google.com/actions/identity/google-sign-in


omenocal:
@armonge done. Also replaced some places where it said: `js::method` to `js::function` as this was having issues when being created with sphinx

src/VoxaEvent.ts
```diff
@@ -87,6 +90,44 @@ export abstract class VoxaEvent implements IVoxaEvent {
     this.initLogger(logOptions);
   }
 
+  public async getUserInformation(): Promise<any> {
+    if (_.has(this, "google")) {
```
armonge: why not `this.platform.name === 'dialogflow'` ?

src/VoxaEvent.ts
```diff
@@ -87,6 +90,44 @@ export abstract class VoxaEvent implements IVoxaEvent {
...
+      return this.getUserInformationWithGoogle();
+    }
+
+    return await this.getUserInformationWithLWA();
+  }
+
+  public getUserInformationWithGoogle(): any {
```
armonge: please specify the the return type, create an interface if none is available

src/VoxaEvent.ts
```diff
@@ -87,6 +90,44 @@ export abstract class VoxaEvent implements IVoxaEvent {
...
+    const result: any = jwt.decode(this.user.idToken);
```
armonge: `this.user.idToken` ??

src/VoxaEvent.ts
```diff
@@ -87,6 +90,44 @@ export abstract class VoxaEvent implements IVoxaEvent {
...
+    result.givenName = result.given_name;
+
+    delete result.email_verified;
+    delete result.family_name;
+    delete result.given_name;
+
+    return result;
+  }
+
+  public async getUserInformationWithLWA(): Promise<any> {
```
armonge: please create an interface for this return type, would be great if both methods share that BTW

src/VoxaEvent.ts
```diff
@@ -87,6 +90,44 @@ export abstract class VoxaEvent implements IVoxaEvent {
...
+    const httpOptions: any = {
+      json: true,
+      method: "GET",
+      uri: `https://api.amazon.com/user/profile?access_token=${this.user.accessToken}`,
```
armonge: what happens if the user has no accessToken and the developer calls this method? should it return a RuntimeException or something like that?

src/VoxaEvent.ts
```diff
@@ -87,6 +90,44 @@ export abstract class VoxaEvent implements IVoxaEvent {
...
+    };
+
+    const result: any = await rp(httpOptions);
```
armonge: create an interface for what this http call returns

src/VoxaEvent.ts
omenocal: It's the variable that has the JSON Web Token with the user's Google information. It's not an accessToken

armonge: yes but, from the dialogflow implementation for the user object, seems like this is not correct https://github.com/VoxaAI/voxa/blob/v3/src/platforms/dialogflow/DialogFlowEvent.ts#L91

perhaps you're looking for `voxaEvent.dialogflow.conv.user.idToken` ?

omenocal:
@armonge done

src/index.ts
```diff
@@ -7,7 +7,7 @@
 
 export { Tell, Say, SayP, Ask, Reprompt } from "./directives";
 export { VoxaPlatform } from "./platforms";
-export { IVoxaEvent, IVoxaIntentEvent, IVoxaIntent } from "./VoxaEvent";
+export { IVoxaEvent, IVoxaIntentEvent, IVoxaIntent, VoxaEvent } from "./VoxaEvent";
```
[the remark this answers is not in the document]
omenocal: @armonge done

test/alexa/AlexaEvent.spec.ts
```diff
@@ -69,3 +73,57 @@ describe("AlexaEvent", () => {
...
+    const userDetails = await alexaEvent.getUserInformation();
+    expect(userDetails).to.deep.equal(lwaResult);
+  });
+
+  it("should throw an error when accessToken is empty", async () => {
+    const rawEvent = rb.getLaunchRequest();
+    const alexaEvent = new AlexaEvent(rawEvent) as VoxaEvent;
+    alexaEvent.platform = alexaSkill;
+
+    try {
```
armonge: maybe also add something to actually make sure the exception is thrown? right now we're only testing that if the exception is thrown the message is what we expect


```suggestion
    let exceptionWasThrown = false;
    try {
      await alexaEvent.getUserInformation();
    } catch (err) {
      exceptionWasThrown = true;
      expect(err.message).to.equal("this.user.accessToken is empty");
    }
    expect(exceptionWasThrown).to.be.true;
```

omenocal: @armonge done

docs/google-sign-in.rst
```diff
@@ -0,0 +1,28 @@
...
+To perform account linking with Google Sign-In, you ask the user to give consent to access their Google profile. You then use the information in their profile, for example their email address, to identify the user in your system. Check out this `link <https://developers.google.com/actions/identity/google-sign-in>`_ for more information.
+
+With Voxa, you can ask for the user's full name like this:
+
+.. code-block:: javascript
+
+  app.onIntent('ProfileIntent', async (dialogFlowEvent: DialogFlowEvent) => {
+    const userInfo = await dialogFlowEvent.getUserInformation();
+
+    dialogFlowEvent.model.email = userInfo.email;
```
rmberrios: is there any chance that is values are empty? if so do it should be specified in here.

docs/voxa-event.rst
```diff
@@ -77,6 +77,74 @@ The ``voxaEvent`` Object
 
     :returns Array: A string array of the platform's supported interfaces
 
+  .. js:function:: getUserInformation()
+
+    Object with user personal information from the platform being used.
+
+    .. code-block:: json
```
rmberrios: change the codeblock to javascript, comments are highlighted as an error

docs/google-sign-in.rst
omenocal: I'm not totally sure, I have made some tests and I do get values in those variables

src/platforms/VoxaPlatform.ts
```diff
@@ -52,6 +52,10 @@ export interface IVoxaPlatformConfig {
   [key: string]: any;
 }
 
+export interface IDialogFlowPlatformConfig extends IVoxaPlatformConfig {
```
armonge: let's move this platform specific declaration to the dialog flow module

src/platforms/dialogflow/DialogFlowEvent.ts
```diff
@@ -26,6 +26,8 @@ import {
 } from "actions-on-google";
 import { Context as AWSLambdaContext } from "aws-lambda";
 import { Context as AzureContext } from "azure-functions-ts-essentials";
+import { OAuth2Client } from "google-auth-library";
```
armonge: we should add this to the package.json requirements

omenocal: @armonge done

PR #169, opened 2018-11-16
Error classes now are implemented instead of extended

omenocal:


armonge:
this has a lot more commits than it should, please recreate your v3 branch from upstream and then add your changes there

PR #170, opened 2018-11-16
Fixes onUnhandledState not being called

omenocal:
Error classes now are implemented instead of extended. This line
https://github.com/VoxaAI/voxa/blob/v3/src/StateMachine/StateMachine.ts#L145

was not working when the `UnknownState` Error was thrown


armonge:
Let's please also add a test that exercises the previous faulty behavior

PR #173, opened 2018-11-20
Integrate APL directives

wuelcas:
This PR is basically to send the APL directives just like we send the RenderTemplate directives. We can pass the directive object directly into the APL directive key or we can use a variable set inside the views file.

Also added a event that fires when a user touch a TouchWrapper component from a APL Template.

### DIFF
--- a/docs/alexa-directives.rst
+++ b/docs/alexa-directives.rst
@@ -74,7 +74,7 @@ RenderTemplate
 
 `Alexa Documentation <https://developer.amazon.com/docs/custom-skills/display-interface-reference.html>`_
 
-Voxa provides a `DisplayTemplate` builder that can be used with the `alexaRenderTemplate` controller key to create Displa templates for the echo show and echo spot.
+Voxa provides a `DisplayTemplate` builder that can be used with the `alexaRenderTemplate` controller key to create Display templates for the echo show and echo spot.
 
 .. code-block:: javascript
 
--- a/docs/alexa-directives.rst
+++ b/docs/alexa-directives.rst
@@ -95,6 +95,121 @@ Voxa provides a `DisplayTemplate` builder that can be used with the `alexaRender
     });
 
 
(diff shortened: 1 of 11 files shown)


src/platforms/alexa/directives.ts
```diff
@@ -229,6 +229,94 @@ export class RenderTemplate extends AlexaDirective implements IDirective {
...
+
+    if (this.viewPath) {
+      this.directive = await event.renderer.renderPath(this.viewPath, event);
+    }
+
+    this.addDirective(reply);
+  }
+
+  private validateReply(reply: IVoxaReply) {
+    if (reply.hasDirective("Alexa.Presentation.APL.RenderDocument")) {
```
armonge: is this the only validation? can we send an old school render template long with this?

wuelcas: We can, the last directive in the directives array will be the one being rendered. I think just changing the position in the directiveHandlers array will be enough to make the APL template the default in case there's a RenderTemplate in the same response. If the device doesn't support APL but it does support RenderTemplate then the skill will inly send and render the RenderTemplate. I'll make that change and some tests for those cases.

docs/alexa-directives.rst
```diff
@@ -95,6 +95,121 @@ Voxa provides a `DisplayTemplate` builder that can be used with the `alexaRender
...
+          datasources: {},
+          document: {},
+          token: "SkillTemplateToken",
+          type: "Alexa.Presentation.APL.RenderDocument",
+        },
+      };
+    });
+
+
+Alexa Presentation Language (APL) Commands
```
armonge: would be great to also mention the XML tool we use to build this stuff

armonge: Let's keep it separate for now, if just so we can work on it separately from the main voxa repo. But let's also make it a public one. For now, i'm approving this PR

PR #175, opened 2018-11-23
Task/reply keys in transition

armonge:


test/VoxaApp.spec.ts
```diff
@@ -871,5 +882,31 @@ describe("VoxaApp", () => {
...
+          voxaEvent.model.transition = transition;
+        },
+      );
+      const platform = new AlexaPlatform(voxaApp);
+      const response = await platform.execute(launchEvent);
+      expect(response.sessionAttributes.model.transition).to.deep.equal({
+        flow: "yield",
+        reply: "Reply.Say",
+        reprompt: "Reply.Say.reprompt",
+        say: "Reply.Say.say",
```
rmberrios: do you duplicate the say and reply for old-version-support?

PR #176, opened 2018-11-30
Adding LinkOutSuggestion support

armonge:
- Adds LinkOutSuggestion directive for DialogFlow
- Adds documentation
- Updates dependencies


docs/dialogflow-directives.rst
```diff
@@ -147,6 +147,18 @@ Whenever possible, you should incorporate one key suggestion as part of the chat
   });
 
 
+.. code-block:: javascript
+
+  app.onState('someState', () => {
+    return {
+      dialogFlowLinkOutSuggestion: {
+        name: "Suggestion Link",
+        url: 'https://assistant.google.com/',
```
lvelasquezm: @armonge I'm looking into the Actions on Google docs and it says that the `url` property is deprecated and instead we should be using the `openUrlAction` property, I know sending the `url` property may work but I guess we shouldn't be using deprecated things.

https://developers.google.com/actions/reference/rest/Shared.Types/AppResponse#linkoutsuggestion
https://actions-on-google.github.io/actions-on-google-nodejs/interfaces/actionssdk_api_v2.googleactionsv2uielementslinkoutsuggestion.html#url

PR #177, opened 2018-11-30
Adding displayText responses to google actions support

armonge:


wuelcas:
So this new `text` and `textp` are for Google actions only right? are they to display a text in a screen?

armonge:
@wuelcas they worked for botframewrk before, i'm just making it available to google and documenting it now

PR #182, opened 2018-12-17
Added Skill Messaging API

omenocal:
Official documentation here:
https://developer.amazon.com/docs/smapi/skill-messaging-api-reference.html

The methods listed in this class are meant to work out-of-session. That's why I didn't make it extend from the `ApiBase` class like the other api classes. The types for the response for both methods are not listed in the latest version of the Alexa typescript package.

### DIFF
--- a/docs/alexa-apis.rst
+++ b/docs/alexa-apis.rst
@@ -787,11 +787,17 @@ The Skill Messaging API is used to send message requests to skills. These method
 
     Sends message to a skill
 
-    :param endpoint: User's endpoint.
-    :param userId: User's userId.
-    :param data: Object with key-value pairs to send to the skill.
-    :param skillMessagingToken: User's accessToken.
-    :param expiresAfterSeconds: Expiration time in milliseconds, defaults to 3600 milliseconds.
+    :param request: Message request params with the following structure:
+
+  .. code-block:: javascript
+
+    {
+      endpoint: string, // User's endpoint.
+      userId: string, // User's userId.
+      data: any, // Object with key-value pairs to send to the skill.
+      skillMessagingToken: string, // User's accessToken.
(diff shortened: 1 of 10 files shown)


docs/alexa-apis.rst
```diff
@@ -753,3 +753,114 @@ With Voxa, you can create, update, delete and get reminders like this:
...
+    });
+
+    callback(undefined, "OK");
+  };
+
+This will dispatch a 'Messaging.MessageReceived' request to every user and you can handle the code in Voxa like this:
+
+.. code-block:: javascript
+
+  app.onIntent('Messaging.MessageReceived', (voxaEvent) => {
```
armonge: so, why are we treating this as intents? seems like this would not go through the State Machine and has no access to sessionAttributes

docs/alexa-apis.rst
```diff
@@ -753,3 +753,114 @@ With Voxa, you can create, update, delete and get reminders like this:
...
+    const dataReceived = voxaEvent.rawEvent.request.message;
+
+    // DO SOMETHING WITH THE DATA RECEIVED AND RESPOND TO THE REQUEST
```
armonge: also would be great if we could include an example of what can this actually be used for

package.json
```diff
@@ -50,6 +50,7 @@
     "i18next": "^12.1.0",
     "lambda-log": "^2.0.0",
     "lodash": "^4.13.1",
+    "querystring": "^0.2.0",
```
armonge: this module is available inside core node https://devdocs.io/node~8_lts/querystring#querystring_querystring_stringify_obj_sep_eq_options

omenocal:
@armonge done!

PR #187, opened 2019-01-03
Added Facebook directives for Dialogflow

omenocal:


docs/dialogflow-platforms.rst
```diff
@@ -0,0 +1,54 @@
...
+
+.. _facebook:
+
+Facebook Messenger
+==================
+
+The ``DialogflowPlatform`` for voxa has available some of the core functionalities to send to your chatbot in responses. For now, you can integrate:
+
+- Account Linking buttons
+You need to include in your controller the following field: ``facebookAccountLink``, which takes a URL to go into the account linking flow.
```
armonge: please also include a reference to the related facebook documentation

docs/dialogflow-platforms.rst
```diff
@@ -0,0 +1,54 @@
...
+.. code-block:: javascript
+
+  app.onState('someState', () => {
+    return {
+      facebookAccountLink: "https://www.messenger.com"
+    }
+  });
+
+- Postbacks buttons (Suggestion chips)
+You need to include in your controller the following field: ``facebookSuggestionChips``, which could be a simple string that the Voxa renderer will get from your views file with an array of strings, or directly an array of strings.
```
armonge: please also include a link to the related facebook documentation

src/platforms/dialogflow/DialogFlowReply.ts
```diff
@@ -33,6 +33,12 @@ import { addToSSML, addToText, IVoxaReply } from "../../VoxaReply";
 import { DialogFlowEvent } from "./DialogFlowEvent";
 
 export interface IDialogFlowPayload {
+  facebook?: {
+    attachment: {
+      payload: any;
+      type: string;
+    };
+  };
```
armonge: this now makes me think if google should also be an optional property

src/platforms/dialogflow/directives/facebook.ts
```diff
@@ -0,0 +1,85 @@
+import * as _ from "lodash";
+
+import { IDirective } from "../../../directives";
+import { ITransition } from "../../../StateMachine";
+import { IVoxaEvent } from "../../../VoxaEvent";
+import { IVoxaReply } from "../../../VoxaReply";
+import { DialogFlowEvent } from "../DialogFlowEvent";
+import { DialogFlowReply } from "../DialogFlowReply";
+
+export class FacebookAccountLink implements IDirective {
```
armonge: can you check if this works when using the `reply` way of specifying a response?

src/platforms/dialogflow/DialogFlowReply.ts
omenocal: I think it should be, but there are several parts that were breaking when I set it optional that I didn't want to spend more time on fixing, at least not now due to project pressure.

PR #189, opened 2019-01-04
Fixed facebook userId issue

omenocal:


src/platforms/dialogflow/DialogFlowEvent.ts
```diff
@@ -40,6 +40,7 @@ export class DialogFlowEvent extends VoxaEvent {
   public session!: DialogFlowSession;
   public google!: { conv: DialogflowConversation };
   public intent: DialogFlowIntent;
+  public source: string|undefined;
```
armonge: looking at the actual implementation i don't think this can be undefined

src/platforms/dialogflow/DialogFlowEvent.ts
```diff
@@ -84,14 +85,19 @@ export class DialogFlowEvent extends VoxaEvent {
...
     let userId: string;
 
+    this.source = _.get(originalDetectIntentRequest, "source", "google");
+
     if (_.get(storage, "voxa.userId")) {
       userId = storage.voxa.userId;
     } else if (conv.user.id) {
       userId = conv.user.id;
+    } else if (this.source === "facebook") {
+      userId = _.get(originalDetectIntentRequest, "payload.data.sender.id", v1());
```
armonge: Seems like this code right here is the reason you marked `userId` as possibly undefined, i think the following should fix that 

```suggestion
      userId = _.get(originalDetectIntentRequest, "payload.data.sender.id") || v1();
```

src/platforms/dialogflow/DialogFlowEvent.ts
```diff
@@ -84,14 +85,19 @@ export class DialogFlowEvent extends VoxaEvent {
...
     } else {
       userId = v1();
```
armonge: now that i think about it, is this even correct? seems like it will generate a new userId on every request, is not stable

test/dialogflow/directives.spec.ts
```diff
@@ -1047,15 +1047,24 @@ describe("DialogFlow Directives", () => {
...
+          to: "entry",
+        };
       });
 
+      event = _.cloneDeep(require("../requests/dialogflow/facebookLaunchIntent.json"));
+
       const reply = await dialogFlowAgent.execute(event);
+      const attributesJson = JSON.parse(reply.outputContexts[0].parameters.attributes);
+
+      expect(attributesJson.model.userId).to.equal("1234567890");
```
armonge: this is a completely different test, please create it separately, you're testing wether we're adding the user id, but the test says 'should add a facebook account link card'

test/dialogflow/directives.spec.ts
```diff
@@ -1071,13 +1080,22 @@ describe("DialogFlow Directives", () => {
...
```
armonge: same thing here, the expectation is not related to the unit being tested

src/platforms/dialogflow/DialogFlowEvent.ts
omenocal: That's what I was getting when I was testing, but not sure how much would affect to google assistant if I removed this, what do you think?

armonge: i don't think it will affect google assistant as there's code in place to make sure we always get a userId if needed

src/platforms/dialogflow/DialogFlowEvent.ts
```diff
@@ -84,16 +85,19 @@ export class DialogFlowEvent extends VoxaEvent {
    * After that we'll default to the userStorage value
    */
   protected initUser(): void {
+    const { originalDetectIntentRequest } = this.rawEvent;
     const { conv } = this.google;
     const storage = conv.user.storage as any;
-    let userId: string;
+    let userId: string = "";
+
+    this.source = _.get(originalDetectIntentRequest, "source") || "google";
```
armonge: let's initialize this in the constructor

test/dialogflow/DialogFlowEvent.spec.ts
```diff
@@ -177,6 +177,36 @@ describe("DialogFlowEvent", () => {
   });
 });
 
+describe("Facebook Messenger", () => {
```
armonge: please rework this test, seems more like a test for the entire voxa flow than for just the DialogflowEvent, compare it to the other tests in the same module, most of them just initialize the object and then make expectations on the object values/methods

https://github.com/VoxaAI/voxa/blob/1e39c38a964bda48bf29f58657b9e27763e4181d/test/dialogflow/DialogFlowEvent.spec.ts#L105

PR #191, opened 2019-01-09
Version 3

armonge:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "voxa",
-  "version": "3.0.0-alpha42",
+  "version": "3.0.0",
   "description": "A fsm (state machine) framework for Alexa apps using Node.js",
```
lvelasquezm: I guess this description should change, now Voxa is more than only for Alexa 💪

PR #192, opened 2019-01-09
Added Facebook unlink directive

omenocal:
I'm also modifying the variable `source`'s type in `DialogFlowEvent.ts` class from `string` to `string|undefined` because when it's compiled into Javascript, the compiler creates this line in the constructor:

this.source = "";

Since I can't initialize it in the constructor I'm having troubles when this variable gets reseted. This change, fixes that. Test included.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "voxa",
-  "version": "3.0.0",
+  "version": "3.0.0-alpha43",
```
armonge: please remove this

omenocal: Hmm why? I'm planning to release this to NPM to the alpha channel

armonge: So, once we released 3.0.0 the next alpha should be 3.0.1-alpha or 3.1.0-alpha depending on semver. Seems like this would 3.1.0 because there's a new feature

omenocal: Got it, thanks!

PR #193, opened 2019-01-10
Added QuickReplies Facebook directives

omenocal:
Official documentation here:
https://developers.facebook.com/docs/messenger-platform/send-messages/quick-replies

With these quick replies you can send a button to request:

- Location
- PhoneNumber
- Email
- Text Options. Similar to suggestion chips, but these ones disappear after they're clicked.


src/platforms/dialogflow/directives/facebook.ts
```diff
@@ -112,3 +112,119 @@ export class FacebookSuggestionChips implements IDirective {
...
+  constructor(public message: string) {}
+
+  public async writeToReply(
+    reply: IVoxaReply,
+    event: IVoxaEvent,
+    transition: ITransition,
+  ): Promise<void> {
+    const dialogFlowReply = (reply as DialogFlowReply);
+
+    dialogFlowReply.source = "facebook";
```
armonge: seems like most of this directives are similar enough that we can create a helper for that, something like https://github.com/VoxaAI/voxa/blob/master/src/platforms/dialogflow/directives/assistant.ts#L64

Please explora that optino

src/renderers/Renderer.ts
```diff
@@ -70,6 +70,7 @@ export class Renderer {
     const platform = voxaEvent.platform.name;
 
     let message = voxaEvent.t(view, {
+      defaultValue: view,
```
omenocal: When sending a `string` to the key for the directives, Voxa thinks is a path to be rendered by i18n. The value returned is super random since i18n is really trying to render something. When setting the `defaultValue` flag, i18n will return the object that it couldn't translate, and that's exactly what we need for the directives in case the value sent is a string.

@armonge

PR #198, opened 2019-01-29
Added button and OpenGraph templates

omenocal:
- Button template: https://developers.facebook.com/docs/messenger-platform/send-messages/template/button
- OpenGraph template: https://developers.facebook.com/docs/messenger-platform/send-messages/template/open-graph


docs/dialogflow-platforms.rst
```diff
@@ -527,6 +527,185 @@ Or you can also handle these values from your views file
...
+  app.onState('someState', () => {
+    const buttonBuilder1 = new FacebookButtonTemplateBuilder();
+    const buttonBuilder2 = new FacebookButtonTemplateBuilder();
+    const buttonBuilder3 = new FacebookButtonTemplateBuilder();
+    const facebookTemplateBuilder = new FacebookTemplateBuilder();
+
+    buttonBuilder1
+      .setPayload("payload")
+      .setTitle("View More")
+      .setType("postback");
```
lvelasquezm: I think for the `type` property you should create an enum in Voxa's code. I can see there are 9 [types of buttons](https://developers.facebook.com/docs/messenger-platform/send-messages/buttons) users can use and they better go to Voxa docs instead of Facebook docs to see which button type they can use.

PR #200, opened 2019-02-04
Split Dialogflow platforms

omenocal:
Even though the Google Assistant classes cover a lot of methods provided by Dialogflow that can be used by other platforms, it's necessary to separate the logic into their own classes for the different Dialogflow integrations, specially now that we're expanding the framework and every platform has their own built-in interfaces. For example: Google Assistant cards have a different syntax that Facebook cards.

I have left the initial Dialogflow classes, in case someone needs to implement another Dialogflow integrations different thatn Google Assistant and Facebook, like Telegram.

In case there's already a Google Assistant project and you want to update Voxa, you'll simply need to change your imports from `Dialogflow...` to `GoogleAssistant...`, for example:
`GoogleAssistantPlatform` instead of `DialogflowPlatform`.


src/platforms/dialogflow/DialogflowEvent.ts
```diff
@@ -61,6 +61,10 @@ export class DialogflowEvent extends VoxaEvent {
     this.intent = new DialogflowIntent(this.google.conv);
   }
 
+  public async getUserInformation(): Promise<any> {
+    return undefined;
+  }
+
   public async verifyProfile(): Promise<TokenPayload | undefined> {
```
armonge: is this maybe more for the google assistant?

src/platforms/dialogflow/facebook/FacebookPlatform.ts
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright (c) 2018 Rain Agency <contact@rain.agency>
```
armonge: new files get 2019 :)

src/platforms/dialogflow/facebook/FacebookReply.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  constructor(event: FacebookEvent) {
+    this.payload = {
+      facebook: {},
+    };
+
+    this.outputContexts = event.session.contexts;
+  }
+
+  public async saveSession(attributes: IBag, event: IVoxaEvent): Promise<void> {
```
armonge: should this be done in the parent class?

src/platforms/dialogflow/facebook/FacebookReply.ts
```diff
@@ -0,0 +1,119 @@
...
+      this.fulfillmentText = addToText(this.fulfillmentText, statement);
+      this.payload.facebook.text = this.fulfillmentText;
+    }
+  }
+
+  public addReprompt(reprompt: string) {
+    // FACEBOOK MESSENGER DOES NOT USE REPROMPTS
+  }
+
+  public hasDirective(type: string | RegExp): boolean {
```
armonge: this is not about surfaces, is about wether the request includes some directive, say buttons or something like that

src/platforms/dialogflow/facebook/FacebookReply.ts
```diff
@@ -0,0 +1,119 @@
...
+    // FACEBOOK MESSENGER DOES NOT USE SURFACES
+    return false;
+  }
+
+  public get hasDirectives(): boolean {
```
armonge: this one should check if the reply has any directive at all

src/platforms/dialogflow/google/GoogleAssistantEvent.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    this.request = {
+      locale: _.get(rawEvent.queryResult, "languageCode") || "",
+      type: "IntentRequest",
+    };
+
+    this.intent = new GoogleAssistantIntent(this.google.conv);
+  }
+
+  public async verifyProfile(): Promise<TokenPayload | undefined> {
```
armonge: duplicated code, remove it from the parent class

src/platforms/dialogflow/google/GoogleAssistantSession.ts
```diff
@@ -0,0 +1,36 @@
+import {
+  Context,
+  Contexts,
+  DialogflowConversation,
+  Parameters,
+} from "actions-on-google";
+import * as _ from "lodash";
+import { IBag, IVoxaSession } from "../../../VoxaEvent";
+
+export class DialogflowSession implements IVoxaSession {
```
armonge: any reason to not just use that? why do we need specific session classes for facebook and assistant?

armonge:
Seems like the tests are failing in the integration test, you can run all those locally with `./run-ci.sh`

Also, have you tested this with telegram? i don't want it to stop working after we merge this

src/platforms/dialogflow/google/GoogleAssistantSession.ts
omenocal: The `GoogleAssistantSession` class uses objects from the `actions-on-google` package. While `FacebookSession` simply controls sessions in one of the contexts of Dialogflow request, which is what the action-on-google package does as well, but I just didn't want to mix things up between them. I think that maybe I can leave the FacebookSession code in DialogflowSession to be used not only by Facebook but also for any other generic platform like Telegram. And since GoogleAssistant has its own manager, I can make it implement the DialogflowSession class and overwrite that method.

Thoughts on that ?

src/platforms/dialogflow/facebook/FacebookReply.ts
omenocal: I'm not inheriting from `DialogflowReply` but directly from `IVoxaReply`. Also the ways Facebook and GA save the session are different because GA uses the methods from the actions-on-google package, and I didn't want to mix things up with Facebook

omenocal:
I didn't test it, but, I didn't touch the `Dialogflow` classes that the Telegram integration should use. I copied all the code from those clases and created the GoogleAssistant clases, because Dialogflow classes were created thinking about GA. Then, I created Facebook classes and adjusted methods according to the platform.

So, Telegram integration should work because the Dialogflow classes were not modified and are still exposed through the Voxa package. But right now I'm focusing more on Facebook and GA to work, in a future PR I can look and add more work to Telegram

src/directives.ts
```diff
@@ -87,12 +87,12 @@ export class Ask implements IDirective {
...
         reply.addStatement(sampleOrItem(statement, event.platform));
       } else {
-        this.addSttementToReply(statement, reply, event);
+        this.addStatementToReply(statement, reply, event);
       }
     }
   }
 
-  private addSttementToReply(
+  private addStatementToReply(
```
armonge: good catch

src/platforms/alexa/AlexaEvent.ts
```diff
@@ -30,7 +30,9 @@ import { Context as AWSLambdaContext } from "aws-lambda";
 import { Context as AzureContext } from "azure-functions-ts-essentials";
 import { LambdaLogOptions } from "lambda-log";
 import * as _ from "lodash";
-import { IVoxaIntent, VoxaEvent } from "../../VoxaEvent";
+import * as rp from "request-promise";
```
armonge: i just realized recently that the correct way to import rp with typescript is `import rp = require('request-promise');`

at least according to https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/request-promise/request-promise-tests.ts#L1

Let's update it in this file since we're already working on this code

src/platforms/botframework/BotFrameworkEvent.ts
```diff
@@ -64,6 +73,11 @@ export class BotFrameworkEvent extends VoxaEvent {
     this.initUser();
   }
 
+  public async getUserInformation(): Promise<any> {
```
armonge: not sure what's this method? :thinking:

src/platforms/dialogflow/DialogflowEvent.ts
```diff
@@ -35,16 +35,15 @@ import { VoxaEvent } from "../../VoxaEvent";
...
-export interface IGoogle {
+export interface IDialogflow {
   conv: DialogflowConversation;
 }
 
 export class DialogflowEvent extends VoxaEvent {
   public rawEvent!: GoogleCloudDialogflowV2WebhookRequest;
   public session!: DialogflowSession;
-  public google!: IGoogle;
+  public dialogflow!: IDialogflow;
```
armonge: not sure about this, maybe we should make `dialogflow` key an alias of `google` and deprecate the `google` key? 
Otherwise we're making it so projects with an older voxa version will have their code not work anymore

src/platforms/dialogflow/DialogflowEvent.ts
```diff
@@ -95,26 +86,20 @@ export class DialogflowEvent extends VoxaEvent {
...
-      } else {
-        userId = v1();
-      }
-
-      _.set(this.google.conv.user.storage, "voxa.userId", userId);
-    } else if (this.source === "facebook") {
-      userId = _.get(originalDetectIntentRequest, "payload.data.sender.id");
+    if (conv.user.id) {
+      userId = conv.user.id;
+    } else if (_.get(storage, "voxa.userId")) {
```
armonge: i believe the storage is something that's only available on the google actions platform, and so is most of the content in the `conv.user`  object :thinking:

src/platforms/dialogflow/facebook/FacebookReply.ts
```diff
@@ -0,0 +1,117 @@
...
+    attachment?: {
+      payload: any;
+      type: string;
+    };
+    quick_replies?: any[];
+    text?: any;
+  };
+}
+
+export class FacebookReply extends DialogflowReply {
```
armonge: override the payload type to make it `public payload: IFacebookPayload`

src/platforms/botframework/BotFrameworkEvent.ts
omenocal: I made this method abstract in the parent class VoxaEvent:

`public abstract async getUserInformation(): Promise<IVoxaUserProfile>;`

So, I put it in all the children classes. For `BotFrameworkEvent` is just a placeholder, that's why I left it blank

src/platforms/dialogflow/DialogflowEvent.ts
omenocal: Yes, that's why I'm overwriting this method in the `FacebookEvent` class

src/platforms/dialogflow/facebook/FacebookReply.ts
omenocal: It is public and it's already there:
https://github.com/VoxaAI/voxa/pull/200/files/bd0a4389c665f3e31c62f9de1d97ff4cf8d6d40a#diff-60f2950f06e7c6e0a52248fea00a9287R62

src/platforms/dialogflow/DialogflowEvent.ts
```diff
@@ -58,30 +63,22 @@ export class DialogflowEvent extends VoxaEvent {
...
-    );
-
-    return payload;
+  public async getUserInformation(): Promise<any> {
+    return undefined;
   }
 
   protected initSession(): void {
-    this.google = {
+    this.dialogflow = {
```
armonge: didn't we decide `this.google` was gonna be a deprecated property that  proxied to `this.dialogflow` ? I don't see that

src/platforms/dialogflow/DialogflowEvent.ts
```diff
@@ -95,26 +92,20 @@ export class DialogflowEvent extends VoxaEvent {
...
-      userId = _.get(originalDetectIntentRequest, "payload.data.sender.id");
+    if (conv.user.id) {
+      userId = conv.user.id;
+    } else if (_.get(storage, "voxa.userId")) {
+      userId = storage.voxa.userId;
+    } else {
+      userId = v1();
     }
 
+    _.set(this.dialogflow.conv.user.storage, "voxa.userId", userId);
```
armonge: user storage is only available in google actions, let's move this to the related code to the relevant class

src/platforms/dialogflow/DialogflowEvent.ts
omenocal: @armonge at the top of the class you'll find this
```
// DEPRECATED: INSTEAD USE dialogflow
  get google() {
    return this.dialogflow;
  }
```

src/platforms/dialogflow/facebook/FacebookEvent.ts
```diff
@@ -106,6 +93,21 @@ export class FacebookEvent extends DialogflowEvent {
       userId,
     };
   }
+
+  private getFacebookProfile(userFields: string) {
+    const params = [
```
armonge: instead of adding querystring params like this, it's better to use `qs`

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/request/index.d.ts#L129
https://github.com/request/request#requestoptions-callback

PR #208, opened 2019-02-18
Remove onStateMachineError from docs

lvelasquezm:
The `onStateMachineError` error handler no longer exists in recent versions of Voxa

### DIFF
--- a/docs/voxa-app.rst
+++ b/docs/voxa-app.rst
@@ -164,18 +164,6 @@ You can register many error handlers to be used for the different kind of errors
 
 They're executed sequentially and will stop when the first handler returns a reply.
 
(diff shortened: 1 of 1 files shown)


lvelasquezm:
> Random thought, on open source project they labeled the function as deprecated

That's right, I guess we could label it as deprecated if there is an equivalent handler for the deprecated one, I think that's not the case for `onStateMachineError`? I know there's an `onError` handler but that one already exists from previous versions. Thoughts?

rmberrios:
Yeah , we could tell them to use onError instead

armonge:
deprecated is different, deprecated means it's avaialble, but no longer recommended to use in favor of another option. Not the case here

PR #210, opened 2019-02-27
Adding support for the elicit dialog directive in alexa.

camaronc:


armonge:
Maybe even add some e2e tests, something like https://github.com/VoxaAI/voxa/blob/dev/test/e2e/alexa.spec.ts ?

PR #214, opened 2019-03-11
Added Handover Protocol Facebook helpers

omenocal:
Docs:
https://developers.facebook.com/docs/messenger-platform/handover-protocol

### DIFF
--- a/docs/dialogflow-events.rst
+++ b/docs/dialogflow-events.rst
@@ -48,3 +48,69 @@ The ``FacebookEvent`` Object
   });
 
   const reply = await facebookBot.execute(event);
+
+  The ``facebookEvent`` object also gives you the necessary helpers to implement the Handover Protocol, very useful when you want to pass the conversation from your bot to a live person, the most common example is when user sends to your bot the following text: I want to talk to a representative. This means your bot is not understanding what user is saying or the bot can't give to the user what they are looking for. So, it's necessary a person to talk directly to the user. You can pass the control to your Page Inbox like this:
+
+.. code-block:: javascript
+
+  const { FacebookEvent, FacebookPlatform, VoxaApp } = require('voxa');
+
+  const config = {
+    pageAccessToken: 'EAAaKuJF183EBAApxv.........',
+  };
+  const app = new VoxaApp({ views, variables });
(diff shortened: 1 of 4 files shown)


docs/dialogflow-events.rst
```diff
@@ -48,3 +48,69 @@ The ``FacebookEvent`` Object
...
+    await voxaEvent.facebook.passThreadControlToPageInbox();
+
+    return {
+      flow: "terminate",
+      text: "Facebook.RepresentativeWillGetInTouch.text",
+      to: "die",
+    };
+  });
+
+  Also, if the app you are working on is not the Primary Receiver, you can request control of the conversation like this:
```
armonge: There seems to be some formatting issue in the rst doc, probably one extra tabe missing or something like that. Check how this is rendered in https://github.com/VoxaAI/voxa/blob/bb0daac7c95ccefd49ca5f1186a0726d25d4b28f/docs/dialogflow-events.rst

docs/dialogflow-events.rst
```diff
@@ -48,3 +48,69 @@ The ``FacebookEvent`` Object
...
+    await voxaEvent.facebook.requestThreadControl();
+
+    return {
+      flow: "terminate",
+      text: "Facebook.ControlRequested.text",
+      to: "die",
+    };
+  });
+
+  Finally, if you detect the secondary receiver is not responding to the user, you can make your bot (Primary Receiver) take the control of the conversation like this:
```
armonge: how would the bot detect the secondary receiver is not responding? can we give an example of this? does the bot still get the interactions between the agent and the user?

omenocal: The primary bot needs to be subscribed to the `standby` channel as the docs suggest:
https://developers.facebook.com/docs/messenger-platform/handover-protocol#standby_channel

I'm not explaining all of that, because it's in the docs and the developer should know about it.

I don't have a specific use case in mind, but I come up with something like this:
If after 30 mins of receiving the `standby` event, there are not more events, there could be a scheduled event running in AWS that sends a `ControlIntent` to Dialogflow with user's information so Primary Bot requests/takes the control over the conversation.

Again, I simply wanted to add the method to the Voxa helpers, but I don't have in mind a use case I want to use this for right now

PR #225, opened 2019-05-21
Add docs for StopAudio directive and resuming an audio

wuelcas:


rmberrios:
It will be good to add how to set the metadata for an audio directive.

wuelcas:
Done!

PR #229, opened 2019-06-10
New release 3.1.3

omenocal:


docs/alexa-directives.rst
```diff
@@ -213,27 +213,124 @@ An APL Command is sent with the `alexaAPLCommand` key in your controller. Just l
...
 
-    app.onState('someState', () => {
-      const playAudio = new PlayAudio(
-        'http://example.com/example.mp3',
-        '{}',
-        0,
-        'REPLACE_ALL'
-      );
+        return {
+          directives: [playAudio],
```
armonge: I don't really like the fact that all other directives can be called directly using some key in the transition object, but the PlayAudio one actually requires one to import the directive class. Feels like that's something that should be looked into

src/VoxaApp.ts
```diff
@@ -542,13 +539,13 @@ export class VoxaApp {
...
   public async transformRequest(voxaEvent: IVoxaEvent): Promise<void> {
-    await this.i18nextPromise;
+    await initializeI118n(this.config.views);
     const data = voxaEvent.session.attributes.model as IBag;
     const model = await this.config.Model.deserialize(data, voxaEvent);
 
     voxaEvent.model = model;
     voxaEvent.log.debug("Initialized model ", { model: voxaEvent.model });
-    voxaEvent.t = this.i18n.getFixedT(voxaEvent.request.locale);
+    voxaEvent.t = i18n.getFixedT(voxaEvent.request.locale);
```
armonge: are we sure this is correct? Before this change there was a new instance of i18n created for every request. Seems like this change will make it so there's only one global instance. In the past this was an issue for multi language skills, a german skill would get the english translations from time to time, or viceversa

@omenocal

PR #230, opened 2019-06-10
Initializes i18n instance in every request

omenocal:


src/VoxaApp.ts
```diff
@@ -90,10 +90,13 @@ export class VoxaApp {
 
   public config: IVoxaAppConfig;
   public renderer: Renderer;
+  public i18nextPromise: PromiseLike<i18next.TFunction>;
+  public i18n: i18next.i18n;
   public states: State[] = [];
   public directiveHandlers: IDirectiveClass[];
 
   constructor(config: any) {
+    this.i18n = i18n.createInstance();
```
armonge: probably need some sort of test for the issue this code is trying to fix. To prevent someone from removing it again and even some sort of documentation. Would be great to also add some comment there saying "We initialize a new instance on every request because...."

omenocal: In this case I'm not fixing an issue, I was just updating the classes from the newest version of the i18n library, but I agree about adding a warning message to not remove the variable and put it away as a global variable.

armonge: Also, is this really per request? seems like it's being created once on VoxaApp instantiation :thinking: maybe it's fine like that? Just saying we need some test before even attempting the change cause it might be that it's a none issue? Were you able to reproduce?

armonge: If the main goal of the PR is to update the versions, perhaps it's wrongly titled?

omenocal: I updated the classes in my last PR but left the i18n variable as global. In this PR I'm putting it back as a local variable of the VoxaApp class.

And I do think is per request, and by request I mean, every time the user speaks to Alexa. With that, the Voxa object is initialized too, isn't it ? 🤔

I'm just putting back the code as it was before me messing up with that detail. I couldn't reproduce it, but if it's now back as it was that means the test you created for that issue will cover it, or am I wrong? :D

omenocal: I think I was able to reproduce the error when I leave the variable as global. Will send a push and LMK your thoughts about it

omenocal: @armonge done!

Issue #238, opened 2019-06-27
States with an intents array filter should have priority over the ones that don't


wuelcas:
Suppose you have this code:

```
voxaApp.onState("myExampleState", { ... });
voxaApp.onState("myExampleState", { ... }, "YesIntent");
```

If the user triggers the `YesIntent` and the state in the session is `myExampleState`, the code that will run is the first controller that Voxa founds of that state, in this case is the one without the "YesIntent" filter. For this case, if you don't have a code that handles the `YesIntent` it may land in a fallback or unhandle state, and that leads to frustration.

The order on how we structure our state controllers matter in Voxa and that's great, but in this kind of situation I think Voxa should prioritize the controllers with an intent filter over the ones that don't even if the controller without an intent filter is the first one in the code structure, just because how explicit setting that filter is.

A seasoned Voxa dev can know that controllers with intent filters goes on top of the one that don't, but someone that's getting around Voxa things can't tell that, and it can be frustrating to know why I can't land on the controller that I'm being explicit about, and it's more frustrating that Voxa doesn't throw a clue on why that happens.

PR #239, opened 2019-06-27
Prioritize controllers with an intents array for the same state

wuelcas:
This PR tackles the problem described on #238. Please give it a read and let me know what you think.

The order of how we structure our code still matters in this change, but a controller with an intent filter has more priority over one that don't for the same state no matter the order of the controllers.

So if I have:

```
voxaApp.onState("myExampleState", { ... });
voxaApp.onState("myExampleState", { ... }, "YesIntent");
```

And the user triggers a `YesIntent`, the controller with the `YesIntent` filter is the one being executed.

### DIFF
--- a/src/StateMachine/StateMachine.ts
+++ b/src/StateMachine/StateMachine.ts
@@ -254,8 +254,8 @@ export class StateMachine {
       return states[0];
     }
 
-    // If the code reaches this points that means the `states` array contains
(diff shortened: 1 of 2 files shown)


wuelcas:
I'm realizing that I made this PR to merge to `master` instead of `dev`. Closing this PR and I will be doing the same PR for `dev`. Sorry.

PR #240, opened 2019-06-27
Set not required play audio params in the interface

wuelcas:
The `offsetInMilliseconds`, `behavior` and `metadata` params should be optionals in the interface definition

### DIFF
--- a/src/platforms/alexa/directives.ts
+++ b/src/platforms/alexa/directives.ts
@@ -457,9 +457,9 @@ export class AccountLinkingCard implements IDirective {
 export interface IAlexaPlayAudioDataOptions {
   url: string;
   token: string;
-  offsetInMilliseconds: number;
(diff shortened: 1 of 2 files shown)


armonge:
Do we have tests where our code without those parameters? will our code not crash if a user decides to send a directive like that without those parameters

wuelcas:
It will not crash since later in the directive code, default values are set for those params, like this:

```
this.directive = {
  audioItem: {
    metadata: this.data.metadata || {},
    stream: {
      offsetInMilliseconds: this.data.offsetInMilliseconds || 0,
      token: this.data.token,
      url: this.data.url,
    },
  },
  playBehavior: this.data.behavior || "REPLACE_ALL",
  type: "AudioPlayer.Play",
};
```

But I'll create a test for this.

PR #248, opened 2019-07-08
Set unused params as optionals in directives method

### LINKED ISSUE #247: Directive method requires three parameters when only needs one
Every directive has a `writeToReply` method that requires three parameters: `reply`, `event` and `transition`. Those three params are not used in all directives, most of the directives just use the first one.

When building a skill in vanilla Javascript you can just pass the one you need to pass, but when using Typescript you are obligated to pass the three params.

One practical example for using the `writeToReply` method inside a skill, is when writing the PlayAudio directive in a reply on a `AudioPlayer.PlaybackNearlyFinished` request. That directive just need the `reply` param, the other two params are not used at all.

```
voxaApp['onAudioPlayer.PlaybackNearlyFinished']((voxaEvent: IVoxaEvent, reply: AlexaReply) => {
  const playAudio = new PlayAudio({
    behavior: "REPLACE_ALL",
    offsetInMilliseconds: 0,
    token: "",
    url: 'https://domain.com/audio.mp3'
  });

  // If I pass only the reply param, Typescript will complain
  playAudio.writeToReply(reply, voxaEvent, {});

  return reply;
});
```

wuelcas:
Fixes #247 

So I went through all existing directives and set unused params as optional, not for every directive but for the ones not using the params.

So, using the example in issue #247, I can now just use it in Typescript as: `  playAudio.writeToReply(reply);`

### DIFF
--- a/docs/voxa-app.rst
+++ b/docs/voxa-app.rst
@@ -269,15 +269,16 @@ Handle events from the `AudioPlayer interface <https://developer.amazon.com/publ
   .. code-block:: javascript
 
     app['onAudioPlayer.PlaybackNearlyFinished']((voxaEvent, reply) => {
-      const directives = {
-        type: 'AudioPlayer.Play',
-        playBehavior: 'REPLACE_ENQUEUED',
-        token: "",
-        url: 'https://www.dl-sounds.com/wp-content/uploads/edd/2016/09/Classical-Bed3-preview.mp3',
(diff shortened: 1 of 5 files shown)


armonge:
@wuelcas  Would be really great to also add the sample from the issue as documentation

wuelcas:
@armonge Done!

PR #288, opened 2020-02-19
Entity Directive

sloaisiga:
This Directive creates both: Dynamic Entity and Session Entity based on a generic structure, which depends on the platform.


armonge:
@sloaisiga can you rebase the PR? easiest would be to have a new branch from the remote dev, otherwise the it shows like there's a lot more commits than there really are

src/directives.ts
```diff
@@ -7,8 +7,11 @@
  * return { reply: 'View' }
  */
 
+import { Response } from "ask-sdk-model";
 import * as bluebird from "bluebird";
 import * as _ from "lodash";
+import { DialogflowReply } from ".";
+import { AlexaReply } from "../src/platforms/alexa/AlexaReply";
```
armonge: I'm not sure i quite like this, it's platform specific code but we're implementing it in the core part of the framework

armonge:
I'm not against a change that would allow to use the same structure for both google and alexa, but i would much rather prefer if this was implemented in the specific platforms instead of like this

wuelcas:
Actually I think we have directives for the specific platforms, I understand this is just another option in case the developer doesn't want to do the same thing for both platforms in two separate directives, am I right @sloaisiga ?

sloaisiga:
> I'm not against a change that would allow to use the same structure for both google and alexa, but i would much rather prefer if this was implemented in the specific platforms instead of like this

The idea is to have a generic structure for both.

sloaisiga:
> Actually I think we have directives for the specific platforms, I understand this is just another option in case the developer doesn't want to do the same thing for both platforms in two separate directives, am I right @sloaisiga ?

Yes, you're right. This is another way to implement both directives if needed, but using this "generic" entity directive.

armonge:
I completely understand the case for having a single API with the same structure for both the Alexa and Google entities. What i'm saying is that the code to support this should not be in the parts of voxa that handle the core directives, the code should be implemented in https://github.com/VoxaAI/voxa/tree/master/src/platforms/alexa and https://github.com/VoxaAI/voxa/tree/master/src/platforms/dialogflow not in https://github.com/VoxaAI/voxa/tree/master/src

If you check https://github.com/VoxaAI/voxa/pull/288/files#diff-5d0c7b82b6c8a89bd1aa24e310f21465R17 you'll notice there's a lot of platform specific code and checks like `if platform === 'google'` all of that should be moved away from core and into the platforms

sloaisiga:
> I completely understand the case for having a single API with the same structure for both the Alexa and Google entities. What i'm saying is that the code to support this should not be in the parts of voxa that handle the core directives, the code should be implemented in https://github.com/VoxaAI/voxa/tree/master/src/platforms/alexa and https://github.com/VoxaAI/voxa/tree/master/src/platforms/dialogflow not in https://github.com/VoxaAI/voxa/tree/master/src
> 
> If you check https://github.com/VoxaAI/voxa/pull/288/files#diff-5d0c7b82b6c8a89bd1aa24e310f21465R17 you'll notice there's a lot of platform specific code and checks like `if platform === 'google'` all of that should be moved away from core and into the platforms

@armonge what do you think about creating a new share folder inside platform (src/platforms/share) to include this new Directive?

sloaisiga:
Moved to #289

PR #289, opened 2020-02-21
Entity Directive 

sloaisiga:
This Directive creates both: Dynamic Entity and Session Entity based on a generic structure, which depends on the platform.

### DIFF
--- a/src/platforms/share/Entity.ts
+++ b/src/platforms/share/Entity.ts
@@ -1,4 +1,4 @@
-import { er } from "ask-sdk-model";
+// import { er } from "ask-sdk-model";
 import * as _ from "lodash";
 import { IVoxaEvent } from "../../VoxaEvent";
 
@@ -49,7 +49,7 @@ export abstract class EntityHelper {
           };
         }
         return entity;
-      },
+      }
     );
 
     return {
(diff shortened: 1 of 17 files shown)


src/platforms/share/Entity.ts
```diff
@@ -0,0 +1,279 @@
...
+    let entity: any;
+    let behavior: any;
+
+    ({ entity, behavior } = this.generateEntityFormat(
+      rawEntity,
+      platform,
+      event,
+    ));
+
+    if (platform === "alexa") {
```
armonge: it should be possible to compose our code have this checks in the platform specific functions and not in the share code that's supposed to be platform independent

src/platforms/share/Entity.ts
```diff
@@ -0,0 +1,206 @@
...
+  protected sessionEntityObject(
+    property: any,
+    entityOverrideMode: string,
+    name: string,
+    event: IVoxaEvent,
+  ): any {
+    return {
+      entities: property.entities.map((item: any) => this.entityValues(item)),
+      entityOverrideMode,
+      name: `${event.rawEvent.session}/entityTypes/${name}`,
```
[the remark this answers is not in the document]
sloaisiga: yes, but this method is used in SessionEntity and Entity directives, so I placed it here to avoid duplicated code.

src/platforms/share/Entity.ts
```diff
@@ -0,0 +1,232 @@
...
+    }
+    if (!_.isArray(entity) || _.isEmpty(entity)) {
+      throw new Error(
+        "Please verify your entity it could be empty or is not an array",
+      );
+    }
+    return entity;
+  }
+
+  protected createSessionEntity(
```
armonge: this is ONLY used in dialogflow, move it there

src/platforms/share/Entity.ts
```diff
@@ -0,0 +1,232 @@
...
+
+        filteredEntity.push(newEntity);
+        return filteredEntity;
+      },
+      [],
+    );
+    return entity;
+  }
+
+  protected createDynamicEntity(
```
armonge: this is ONLY used in alexa, move it there

src/platforms/share/Entity.ts
```diff
@@ -0,0 +1,180 @@
+import { er } from "ask-sdk-model";
```
armonge, line 1: again, this is alexa specific

src/platforms/share/Entity.ts
```diff
@@ -0,0 +1,180 @@
...
+    behavior = behavior || defaultBehavior;
+
+    if (!_.includes(behaviorList, behavior)) {
+      throw new Error(error);
+    }
+    return behavior;
+  }
+
+  protected validateAlexaEntityBehavior(property: any): any {
+    let behavior: er.dynamic.UpdateBehavior = _.get(property, "updateBehavior");
```
armonge, line 113: alexa specific

armonge:
There's still platform specific code in the shared functions

src/platforms/share/Entity.ts
```diff
@@ -0,0 +1,149 @@
...
+  Clear = "CLEAR",
+}
+
+export interface ISessionEntity {
+  name: string;
+  entities: IEntity[];
+  entityOverrideMode: EntityOverrideMode;
+}
+
+export enum EntityOverrideMode {
```
armonge, line 15: this are all specific to Dialogflow...

PR #290, opened 2020-04-03
Google telephony 

sloaisiga:
Hi Team, 

This is a  new directive for google platform in order to provide a Telephony rich response messages, which can be 3 types: Play Audio, Synthesize speech and Transfer call.

**Link:** https://cloud.google.com/dialogflow/docs/integrations/phone-gateway#rich

### DIFF
--- a/docs/google-assistant-directives.rst
+++ b/docs/google-assistant-directives.rst
@@ -681,3 +681,68 @@ Create an entity in your dialogflow agent and make sure that define synonyms opt
       sayp: "Hello!",
       to: "entry",
     });
+
+Telephony
+-------------------------------------------
+
+`Google Documentation <https://cloud.google.com/dialogflow/docs/integrations/phone-gateway#rich>`_
+
+The Dialogflow phone gateway feature provides a telephone interface to your agent. It is used to build conversational IVR (interactive voice response) solutions that integrate with the rest of your call center network. Currently, you can select a telephone number hosted by Google. In the future, you will also be able to port an existing telephone number.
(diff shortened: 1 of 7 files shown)


src/platforms/dialogflow/google/GoogleAssistantPlatform.ts
```diff
@@ -51,6 +51,7 @@ import {
   SessionEntity,
   Suggestions,
   Table,
+  Telephony,
```
armonge, line 54: This are not google assistant directives, but instead a whole other platform. Maybe it makes sense to work on it similarly to the way we worked the Dialogflow facebook platform? 

https://github.com/VoxaAI/voxa/tree/master/src/platforms/dialogflow/facebook

docs/google-assistant-directives.rst
```diff
@@ -681,3 +681,68 @@ Create an entity in your dialogflow agent and make sure that define synonyms opt
...
+      telephony: "myTelephony",
+      flow: "yield",
+      sayp: "Hello!",
+      to: "entry",
+    });
+
+  // Or you can do it directly...
+
+    app.onState('someState', {
+      telephony: {
```
armonge, line 742: wouldn't it make more sense to have a `transpherPhoneCall` directive or similar?

sloaisiga: Yes it would make sense to have 1 directive per message type too, but also it would be nice to have another one that can accept the several telephony messages at once.

PR #292, opened 2020-04-30
Pinning botbuilder to 3.16

chrux:
Fixing Error: Cannot find module 'skills-validator' brought by new botbuilder package version.


jlmatus:
@chrux  shouldn't this also change the CHANGELOG.md (since you're fixing a bug)?

